PR #4, opened 2016-01-12
Fixed got unexpected string in guest.

mazhang:
Signed-off-by: Maosheng Zhang <mazhang@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -842,7 +842,7 @@ class Expect(Tail):
         :raise ExpectError: Raised if an unknown error occurs
         """
         def get_last_nonempty_line(cont):
-            nonempty_lines = [l for l in cont.splitlines() if l.strip()]
+            nonempty_lines = [l for l in cont.splitlines() if re.match(r"\S+", l)]
             if nonempty_lines:
                 return nonempty_lines[-1]
             else:


mazhang:
ID 1288367

mazhang:
@lmr please help to check why the Travis CI build failed.

Thanks,
Mazhang.

lmr:
@mazhang, the Travis problem was solved. Please rebase.

lmr:
And by the way, the patch could have a more descriptive commit message. As it is today, it's hard to understand why the fix is needed at all.

lmr:
Whoops, I've closed this by accident. Reopened.

mazhang:
The bug still happened will this patch, so close this pull request.

ldoktor:
Well I tried many strings and it never failed. Anyway I found somewhere, that explicit use of `str()` sometimes helps (it might corrupt the encoding, when unsupported one is used, but it shouldn't fail). @mazhang can you try a version with `nonempty_lines = [l for l in cont.splitlines() if str(l).strip()]`, eventually `nonempty_lines = [l for l in [str(_) for _ in cont.splitlines()] if l.strip()]`?

mazhang:
@ldoktor Thanks for your help, I have tried with `nonempty_lines = [l for l in [str(_) for _ in cont.splitlines()] if l.strip()]`, problem still exist.

This problem is from timedrift.py in tp-qemu need get guest time before and after load to check if guest time drift, sometimes the test will failed with time string unexpected.
```
autotest/client/tests/virt/virttest/utils_test/__init__.py", line 304, in get_time
    raise error.TestError("The time string from guest is unexpected.")
```
The session log is:
2016-01-24 04:21:03: [root@dhcp71-85 ~]#
2016-01-24 04:21:23: date +'TIME: %a %m/%d/%Y %H:%M:%S.%N'     <-- This is the command which script send
2016-01-24 04:21:23: TIME: Sun 01/24/2016 17:21:23.944658664
2016-01-24 04:21:24: 
2016-01-24 04:21:24: 
2016-01-24 04:21:24: [root@dhcp71-85 ~]#
2016-01-24 04:21:24: [root@dhcp71-85 ~]#
2016-01-24 04:21:25: [root@dhcp71-85 ~]#

But the script got following messages:
01/24 04:21:25 DEBUG|  __init__:0303| The time string from guest is:
[root@dhcp71-85 ~]#
[root@dhcp71-85 ~]#

I found when guest cpu on very heavy load, read_up_to_prompt() which in aexpect.py will timeout(~1/50), somehow the exception part of cmd_output_safe() drop the correct return and cause get_time() failed.

I think the problem may be in following part code, but still need investigate how it works.
```
def cmd_output_safe(self, cmd, timeout=60, internal_timeout=None,
                        print_func=None):

...
    while (time.time() - start_time) < timeout:
        try:
            o += self.read_up_to_prompt(0.5)
            success = True
            break
        except ExpectError, e:
            o = self.remove_command_echo(e.output, cmd)
            if isinstance(e, ExpectTimeoutError):
                self.sendline()
            elif isinstance(e, ExpectProcessTerminatedError):
                raise ShellProcessTerminatedError(cmd, e.status, o)
            else:
                raise ShellError(cmd, o)

    if not success:
        raise ShellTimeoutError(cmd, o)
...

```

PR #6, opened 2016-05-07
aexpect/client.py: Add safe mode when execute cmd.

CongLi:
1. Add param "safe" when execute cmd, default is False.
   In serial sessions, frequently the kernel might print debug or
   error messages that make read_up_to_prompt to timeout. Let's try
   to be a little more robust and send a carriage return, to see if
   we can get to the prompt when safe=True.

Signed-off-by: Cong Li <coli@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1022,7 +1022,7 @@ class ShellSession(Expect):
                                                  print_func)[1]
 
     def cmd_output(self, cmd, timeout=60, internal_timeout=None,
-                   print_func=None):
+                   print_func=None, safe=False):
         """
         Send a command and return its output.
 
@@ -1032,6 +1032,11 @@ class ShellSession(Expect):
         :param internal_timeout: The timeout to pass to read_nonblocking
         :param print_func: A function to be used to print the data being read
                 (should take a string parameter)
+        :param safe: Whether using safe mode when execute cmd.
+                In serial sessions, frequently the kernel might print debug or
+                error messages that make read_up_to_prompt to timeout. Let's try
+                to be a little more robust and send a carriage return, to see if
+                we can get to the prompt when safe=True.
 
         :return: The output of cmd
         :raise ShellTimeoutError: Raised if timeout expires
@@ -1039,6 +1044,8 @@ class ShellSession(Expect):
                 terminates while waiting for output
         :raise ShellError: Raised if an unknown error occurs
         """
+        if safe:
+            return self.cmd_output_safe(cmd, timeout)
         logging.debug("Sending command: %s" % cmd)
         self.read_nonblocking(0, timeout)
         self.sendline(cmd)
@@ -1102,7 +1109,7 @@ class ShellSession(Expect):
         return self.remove_last_nonempty_line(self.remove_command_echo(o, cmd))
 
     def cmd_status_output(self, cmd, timeout=60, internal_timeout=None,
-                          print_func=None):
+                          print_func=None, safe=False):
         """
         Send a command and return its exit status and output.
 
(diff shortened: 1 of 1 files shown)


CongSmile:
e.g.
2016-04-27 15:44:07: [root@virtlab-66-85-156 ~]#
2016-04-27 15:44:08: ethtool -L eth0 combined 2
2016-04-27 15:44:08: [root@virtlab-66-85-156 ~]#
2016-04-27 15:44:08: echo $?
2016-04-27 15:44:08: 0
2016-04-27 15:44:08: [root@virtlab-66-85-156 ~]# [   98.914335] TCP: lp registered

xutian:
@ldoktor  would you like to review this PR,  this defeat make many tests in failed specially test case for virtual nic device.
thanks in advance.
Xu

CongLi:
@lmr 
Could you help review this one?
We often meet such problem above, which will blocks our testing.

Thanks.

aexpect/client.py
```diff
@@ -1154,9 +1207,32 @@ def cmd_status(self, cmd, timeout=60, internal_timeout=None,
         :raise ShellStatusError: Raised if the exit status cannot be obtained
         :raise ShellError: Raised if an unknown error occurs
         """
+        if safe:
+            cmd_status_safe(cmd, timeout)
```
ldoktor: This should be probably `return self.cmd_status_safe`.

aexpect/client.py
```diff
@@ -1135,8 +1149,42 @@ def cmd_status_output(self, cmd, timeout=60, internal_timeout=None,
...
+                output is the output of cmd
+        :raise ShellTimeoutError: Raised if timeout expires
+        :raise ShellProcessTerminatedError: Raised if the shell process
+                terminates while waiting for output
+        :raise ShellError: Raised if an unknown error occurs
+        """
+        o = self.cmd_output_safe(cmd, timeout)
+        try:
+            # Send the 'echo $?' (or equivalent) command to get the exit status
+            s = self.cmd_output_safe(self.status_test_command, 10)
```
ldoktor: Well the problem of this is, that it won't work in case there is an output before/after the status:
```
0[3514161.846520] smpboot: Booting Node 0 Processor 7 APIC 0x7
[medic@w541 aexpect [1;31m[0m]$ 
```

anyway this should be quite rare and the only alternative I could see is something I wrote few years ago, but it's not portable to windows: https://github.com/autotest/autotest-docker/pull/192/commits/446ecf2302d20bb063e9653c7b2ebc6e905cd3c2#diff-2f87a33ba3bd086834255190bdff5b75R603

so I think we could go this way for now. I'd include a `:warning: This fails in case extra message is printed on the same line as the status output`, so people understand the limitation.

CongLi: Yea, that's a problem.

Thanks @ldoktor .

ldoktor:
Hello @CongLi, I see the function `cmd_output_safe` was already present, so this basically just adds the `*status*safe` functions and couple of aliases, right?

I was not aware of this function, there are few limitations (and strange code) in it, but it's already there, so let's leave it intact.

Regarding the changes here. I don't like the new `safe` parameter, that's very useful. On the other hand I don't think it's necessary to add the aliases. The `cmd_status_output` could simply call `cmd_output` and pass the `safe` argument to it. There is no need for yet another very similar functions. Would that work for you?

CongLi:
@ldoktor 
You are right, and I totally agree with you.
That way is more clear and concise.
And code has been updated. :)

ldoktor:
Nice, thank you for the changes. Btw today I actually suffered the very same problems, my CI failed, because kernel wrote some garbage over the output.

PR #7, opened 2016-08-07
Py3K syntax fixes

clebergnu:
These are simple Python 3 syntax changes that adds compatibility
(as far as I exercised it) to aexpect.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1051,7 +1051,7 @@ class ShellSession(Expect):
         self.sendline(cmd)
         try:
             o = self.read_up_to_prompt(timeout, internal_timeout, print_func)
-        except ExpectError, e:
+        except ExpectError as e:
             o = self.remove_command_echo(e.output, cmd)
             if isinstance(e, ExpectTimeoutError):
                 raise ShellTimeoutError(cmd, o)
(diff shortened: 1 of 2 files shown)


aexpect/client.py
```diff
@@ -1093,7 +1093,7 @@ def cmd_output_safe(self, cmd, timeout=60):
                 o += self.read_up_to_prompt(0.5)
                 success = True
                 break
-            except ExpectError, e:
+            except ExpectError as e:
```
ldoktor: IIRC this is not compatible with python2.4. I know we officially don't support it in avocado, but we still need to support it in avocado-vt, right?

clebergnu: That's a good point as I don't remember a discussion about aexpect minimum Python version requirements.  At least in Travis, it's configured to test under Python 2.6 and Python 2.7.  @lmr any suggestions here?

Still, how does aexpect influence Avocado-VT (which shares the same Avocado Python version requirements) Python requirements?  I mean, Avocado-VT and Avocado require Python 2.6, so as I see it, there's no link between aexpect + Python 2.4 and Avocado-VT.

ldoktor: I think we should rather ask @will-Do and @xutian. Is it OK to use python2.6 features already or do you still require to stick with the older one? Particularly things like this, are they OK?

lmr:
@clebergnu and @ldoktor, I believe it's safe to throw 2.4 out of the window. We've been dealing with that as a legacy support since 2008, so it's about time we stick to py > 2.6.

aexpect/client.py
ldoktor: OK, I have an ack from @xutian, they use autotest for RHEL5 testing. So let me re-evaluate this PR and merge it...

will-Do: We use python2.6 on RHEL6 and python2.7 on RHEL7, so no problem from us.

PR #11, opened 2016-10-11
aexpect.client: Get encoding or use utf-8 by default

ldoktor:
It's near-impossible to get the correct encoding of the underlying
program as one can execute remote connections or shells with different
environment by aexpect. Let's use PYTHONENCODING or "utf-8" by default
while allowing the user to switch it by changing instance.encoding
variable.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -111,6 +111,8 @@ class Spawn(object):
         self.a_id = a_id or data_factory.generate_random_string(8)
         self.log_file = None
         self.closed = False
+        # Use PYTHONENCODINGS or utf-8 (instead of ascii)
+        self.encoding = os.environ.get("PYTHONENCODING", "utf-8")
 
         base_dir = os.path.join(BASE_DIR, 'aexpect_%s' % self.a_id)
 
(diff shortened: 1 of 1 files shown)


ldoktor:
To check this one can use:

```python
>>> import aepect
>>> aexpect.ShellSession("bash").cmd_output(u"echo ěščřžýáíé")
[medic@w541 avocado ]$ ěščřžýáíé

# or with LANG=cs_CZ.UTF-8
>>> print aexpect.ShellSession("bash").cmd_output("skldafsadfsfd")
[medic@w541 avocado ]$ bash: skldafsadfsfd: příkaz nenalezen

```

which fails without utf-8 support. One can also use different encoding and set it by PYTHONENCODINGS, or just fo the instance.

aexpect/client.py
```diff
@@ -111,6 +111,8 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
         self.a_id = a_id or data_factory.generate_random_string(8)
         self.log_file = None
         self.closed = False
+        # Use PYTHONENCODINGS or utf-8 (instead of ascii)
+        self.encoding = os.environ.get("PYTHONENCODING", "utf-8")
```
lmr: I guess the problem here is that setting instance.encoding is not very visible. I guess we need to create API documentation for aexpect ASAP...

ldoktor: Well ideally no one should need to know this. I mean over the years I saw only one complain related to encoding + from time to time I personally have some issues when running under my native locales. So in 99% of cases the ascii is fine, in 99.9999% of cases the utf-8 is right and the 0.0001% of users should dig deep to solve their issue...

PR #16, opened 2017-01-16
aexpect.client: ignore error when decode

xutian:
If stream of a command (eg. nc connect to a remote socket) is not a using local code page, decode the stream with local code page in strict mode will raise exception like:
 '''UnicodeDecodeError: 'utf8' codec can't decode byte 0xff in position 0: invalid start byte'''.

Signed-off-by: Xu Tian <xutian@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -169,7 +169,7 @@ class Spawn(object):
             sub.stdin.flush()
             # Wait for the server to complete its initialization
             while ("Server %s ready" % self.a_id not in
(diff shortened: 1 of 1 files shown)


lmr:
Sure, I see. Can you add a note to the commit message reporting under which conditions the problem happens? Maybe an existing bug report?

xutian:
@lmr detial detail issue in commit message, can you help to review it.  thanks.

lmr:
@xutian merged, thanks!

PR #17, opened 2017-03-08
RPM packaging: upstream and downstream delta reduction

clebergnu:
This PR is a first step towards reducing the amount of differences between upstream and downstream packages.

The rationale is that these RPM packages are mostly (if not exclusively) developed and used on Fedora on EL, and by having similar packages, downstream can pick updates faster and make them quickly available to users.

This would also greatly help the packaing of projects such as Avocado, in which we currently have to deal with aexpect coming from either our custom repo (and package named `aexpect` or from downstream packages by the name of `python-aexpect`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,11 +43,11 @@ build-deb-all: prepare-source
 
 build-rpm-src: source
 	rpmbuild --define '_topdir %{getenv:PWD}' \
-		 -bs aexpect.spec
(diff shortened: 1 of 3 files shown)


clebergnu:
[Backport Official Fedora SPEC files](https://trello.com/c/uWKfxZHu/887-backport-official-fedora-spec-files)

PR #18, opened 2017-03-21
client: Avoid early finish on ignored chars

ldoktor:
When all received chars are ignored the reader assumes it's over. Let's
simply skip round and wait for really empty input.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -586,9 +586,12 @@ class Tail(Spawn):
                     break
                 if fd in r:
                     # Some data is available; read it
-                    new_data = os.read(fd, 1024).decode(self.encoding, "ignore")
+                    new_data = os.read(fd, 1024)
                     if not new_data:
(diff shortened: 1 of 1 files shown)


aexpect/client.py
```diff
@@ -586,9 +586,12 @@ def print_line(text):
                     break
                 if fd in r:
                     # Some data is available; read it
-                    new_data = os.read(fd, 1024).decode(self.encoding, "ignore")
+                    new_data = os.read(fd, 1024)
```
lmr: Won't this bring back the bug that was fixed in https://github.com/avocado-framework/aexpect/commit/e935dfcfe06be713ff2908743e74b72b8e872d40 ?

lmr: Ah, nevermind.

lmr:
@ldoktor, thanks for the bugfix!

PR #23, opened 2018-03-06
Update to install/use python version specific helper script

mmathesius:
These minor changes should allow both py2 and py3 specific versions of the aexpect helper to be built and installed at the same time without conflicting.

Is there anything external to this package that uses the `aexpect-helper` script?

Would it be OK to simplify this change further and skip having the client code fall back to finding an unversioned helper?

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -17,6 +17,7 @@ API used to run/control interactive processes.
 import time
 import signal
 import os
+import sys
 import re
 import threading
 import shutil
@@ -166,7 +167,15 @@ class Spawn(object):
 
         # Start the server (which runs the command)
         if command:
-            helper_cmd = utils_path.find_command('aexpect-helper')
+            # try to find python specific version of aexpect-helper first, then
+            # try unversioned
+            helper_noversion = 'aexpect-helper'
+            helper_versioned = '{0}-{1.major}.{1.minor}'.format(
+                helper_noversion, sys.version_info)
+            try:
+                helper_cmd = utils_path.find_command(helper_versioned)
+            except utils_path.CmdNotFoundError:
+                helper_cmd = utils_path.find_command(helper_noversion)
             sub = subprocess.Popen([helper_cmd],
                                    shell=True,
                                    stdin=subprocess.PIPE,
--- a/python-aexpect.spec
+++ b/python-aexpect.spec
@@ -34,11 +34,12 @@ sftp, telnet, among others.
 
 %install
 %{__python} setup.py install --root %{buildroot} --skip-build
+mv %{buildroot}%{_bindir}/aexpect-helper %{buildroot}%{_bindir}/aexpect-helper-%{python_version}
 
 %files
 %defattr(-,root,root,-)
 %{python_sitelib}/aexpect*
-%{_bindir}/aexpect-helper
+%{_bindir}/aexpect-helper-*
 
 %changelog
 * Mon Apr 3 2017 Lucas Meneghel Rodrigues <lookkas@gmail.com> - 1.4.0-1


aexpect/client.py
```diff
@@ -166,7 +167,12 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
 
         # Start the server (which runs the command)
         if command:
-            helper_cmd = utils_path.find_command('aexpect-helper')
+            # try to find python specific version of aexpect-helper first, then try unversioned
+            helper_noversion = 'aexpect-helper'
+            helper_versioned = helper_noversion + '-{0.major}.{0.minor}'.format(sys.version_info)
```
clebergnu: A style suggestion (shorter lines, although still larger than 79 chars):

```
            helper_versioned = '{0}-{1.major}.{1.minor}'.format(helper_noversion,
                                                                sys.version_info)
```

aexpect/client.py
```diff
@@ -166,7 +167,12 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
...
+            helper_cmd = utils_path.find_command(cmd_name_versioned, "NOTFOUND")
```
clebergnu: You probably meant:

```python
helper_cmd = utils_path.find_command(helper_versioned, "NOTFOUND")
```

So the obvious fix is `s/cmd_name_versioned/helper_versioned/`, but I suggest you use:

```python
            try:
                helper_cmd = utils_path.find_command(helper_versioned)
            except utils_path.CmdNotFoundError:
                helper_cmd = utils_path.find_command(helper_noversion)
```

python-aexpect.spec
```diff
@@ -34,11 +34,12 @@ sftp, telnet, among others.
 
 %install
 %{__python} setup.py install --root %{buildroot} --skip-build
+mv %{buildroot}%{_bindir}/aexpect-helper %{buildroot}%{_bindir}/aexpect-helper-%{python_version}
```
clebergnu: Looks good, I now understand the remarks on Fedora/EPEL packaging.  Please hold https://src.fedoraproject.org/rpms/python-aexpect/pull-request/1 until we get this fixed.

aexpect/client.py
mmathesius: Updated as suggested, then I let `autopep8 ` decide on the line wrapping...

aexpect/client.py
mmathesius: Good heavens. Yes. How did I accidentally check in that "work in progress" line? Updated as suggested. Thanks!

python-aexpect.spec
mmathesius: Of course. I was planning to hold the PR until a new release of `aexpect` was formally tagged.

clebergnu: It looks like we need more than this... from https://fedoraproject.org/wiki/Packaging:Python#Naming :

"For example, the python3 version of "coverage" must ship executables /usr/bin/coverage-3 and /usr/bin/coverage-3.4 (assuming python3 is currently version 3.4), while the python2 version must provide /usr/bin/coverage, /usr/bin/coverage-2 and /usr/bin/coverage-2.7 (assuming python2 version 2.7)."

mmathesius: I think this particular case falls in a grey area of the packaging guidelines.

As we already discussed, the only thing using the helper script is the library itself--which is now looking for it with the full -X.Y suffix. So, a version of the helper with only the -X suffix would be pointless. Packaging an un-suffixed version of the helper is also pointless, although the packaging guidelines say it must be provided and be for python2.

Of course, the library code could be further updated to also look for the helper with the -X suffix. But the build always creates the library artifacts with full X.Y versioning, so it doesn't seem proper to do that, either.

Would you still like me to add a symlink for the -X version?

PR #24, opened 2018-03-09
Python 3: Update dict.items() and dict.values()

hereischen:
Signed-off-by: Haotong Chen <hachen@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -187,7 +187,7 @@ class Spawn(object):
         self.reader_fds = {}
         try:
             assert(is_file_locked(self.lock_server_running_filename))
-            for reader, filename in self.reader_filenames.items():
+            for reader, filename in list(self.reader_filenames.items()):
                 self.reader_fds[reader] = os.open(filename, os.O_RDONLY)
         except (AssertionError, OSError):
             pass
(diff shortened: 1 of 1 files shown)


aexpect/client.py
```diff
@@ -187,7 +187,7 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
         self.reader_fds = {}
         try:
             assert(is_file_locked(self.lock_server_running_filename))
-            for reader, filename in self.reader_filenames.items():
+            for reader, filename in list(self.reader_filenames.items()):
```
clebergnu: Why is this needed I don't see the reason for turning the `dict_items` (in Python 3) into a `list`. The former even performs better being working as an iterator.

Can you elaborate on the reason for this change?

aexpect/client.py
```diff
@@ -256,7 +256,7 @@ def _close_reader_fds(self):
         """
         Close all reader file descriptors.
         """
-        for fd in self.reader_fds.values():
+        for fd in list(self.reader_fds.values()):
```
clebergnu: Same here.

aexpect/client.py
hereischen: sorry, I was just think this will work for both python 2 and 3. But I was forgetting it is compatible.
Will close this patch. Thanks.

PR #25, opened 2018-03-15
RPM Packaging: sync with Fedora/EPEL

clebergnu:
Let's make upstream SPEC file more in sync with what's on Fedora/EPEL and Makefile similar to Avocado's.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,11 @@ all:
 	@echo "RPM related targets:"
 	@echo "make srpm: Generate a source RPM package (.srpm)"
 	@echo "make rpm: Generate binary RPMs"
+	@echo
+	@echo "Release related targets:"
+	@echo "source-release:     Create source package for the latest tagged release"
+	@echo "srpm-release:       Generate a source RPM package (.srpm) for the latest tagged release"
+	@echo "rpm-release:        Generate binary RPMs for the latest tagged release"
 
 source: clean
 	if test ! -d SOURCES; then mkdir SOURCES; fi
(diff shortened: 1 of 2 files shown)


clebergnu:
[aexpect: sync Fedora/EPEL packaging to upstream](https://trello.com/c/49dN9HG4/1280-aexpect-sync-fedora-epel-packaging-to-upstream)

Makefile
```diff
@@ -40,11 +41,11 @@ build-deb-all: prepare-source
 	# build both source and binary packages
 	dpkg-buildpackage -i -I -rfakeroot
 
-build-rpm-src: source
+srpm: source
 	rpmbuild --define '_topdir %{getenv:PWD}' \
 		 -bs python-aexpect.spec
 
-build-rpm-all: source
+rpm: source
```
ldoktor: ok, just some reordering and change of names.

Makefile
```diff
@@ -16,8 +18,13 @@ all:
...
 
-source:
-	$(PYTHON) setup.py sdist $(COMPILE) --dist-dir=SOURCES
+source: clean
+	if test ! -d SOURCES; then mkdir SOURCES; fi
+	git archive --prefix="aexpect-$(COMMIT)/" -o "SOURCES/aexpect-$(SHORT_COMMIT).tar.gz" HEAD
+
+source-release: clean
+	if test ! -d SOURCES; then mkdir SOURCES; fi
+	git archive --prefix="aexpect-$(VERSION)/" -o "SOURCES/aexpect-$(VERSION).tar.gz" $(VERSION)
```
ldoktor: yep, looks similar to Avocado's Makefile

python-aexpect.spec
```diff
@@ -75,6 +66,7 @@ sftp, telnet, among others.
 - Changed URL to aexpect repo
 - Changed build to use a git archived based source
 - Added released version builds
+- Remove compatiblity with older package name
```
ldoktor: hopefully you're right...

python-aexpect.spec
```diff
@@ -20,6 +20,9 @@
 %global gittar		%{srcname}-%{shortcommit}.tar.gz
 %endif
 
+# Selftests are provided but skipped because they use unsupported tooling.
+%global with_tests 0
```
ldoktor: OK, because inspector is not available the default is 0, do I understand it correctly? Let's hope it'll be added soon so we can enable them.

python-aexpect.spec
```diff
@@ -40,7 +40,11 @@ Source0: https://github.com/avocado-framework/%{srcname}/archive/%{commit}.tar.g
 
 BuildArch: noarch
 Requires: python
-BuildRequires: python, python-setuptools
+BuildRequires: python
+
+%if 0%{?rhel}
+BuildRequires: python-setuptools
```
ldoktor: Well it is required and the fact that on some Fedora versions they are already installed doesn't mean we should not list them (especially when we need conditional to include them on RHEL). I don't think this commit should be included (nor in downstream).

python-aexpect.spec
clebergnu: Yes, that why we have https://trello.com/c/kZbGSE10/1279-inspektor-add-rpm-packaging

python-aexpect.spec
clebergnu: I've actually tried this, and it works as intended.  The RPM build environment guaranteed by `mock` includes a number of packages, including say `gcc`, so there's no need to set explicit (very very common) `BuildRequires`.  The list of packages (or group of packages) change from distro to distro, and from version to version.

So, to make it clear, this is not masked by packages already installed in a given system, it's safe because it will be present in the minimal environment set up by mock when building packages.

python-aexpect.spec
```diff
@@ -59,20 +91,48 @@ sftp, telnet, among others.
...
+ln -s aexpect-helper-%{python2_version} %{buildroot}%{_bindir}/aexpect-helper-2
+
+%if %{with_python3}
+%py3_install
+mv %{buildroot}%{_bindir}/aexpect-helper %{buildroot}%{_bindir}/aexpect-helper-%{python3_version}
+ln -s aexpect-helper-%{python3_version} %{buildroot}%{_bindir}/aexpect-helper-3
+%endif
+
+# use python2 for unversioned executable
+ln -s aexpect-helper-%{python2_version} %{buildroot}%{_bindir}/aexpect-helper
```
ldoktor: That makes `python3-aexpect` unusable without `python2-aexpect`, because `aexpect` only executes `aexpect-helper`. Would you please also add patch to first try the python-specific version first?

Makefile
```diff
@@ -49,12 +51,12 @@ build-deb-all: prepare-source
...
-		 -bs python-aexpect.spec
+	if test ! -d BUILD/SRPM; then mkdir -p BUILD/SRPM; fi
+	mock --old-chroot -r $(MOCK_CONFIG) --resultdir BUILD/SRPM -D "commit $(COMMIT)" -D "commit_date $(COMMIT_DATE)" --buildsrpm --spec python-aexpect.spec --sources SOURCES
 
-rpm: source
-	rpmbuild --define '_topdir %{getenv:PWD}' \
-		 -ba python-aexpect.spec
+rpm: srpm
+	if test ! -d BUILD/RPM; then mkdir -p BUILD/RPM; fi
+	mock --old-chroot -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "commit $(COMMIT)" -D "commit_date $(COMMIT_DATE)" --rebuild BUILD/SRPM/python-aexpect-$(VERSION)-*.src.rpm
```
ldoktor: Are you sure the `--old-chroot` is necessary? I tried it without it on F25 and it worked well. Perhaps it's time to remove this option.

python-aexpect.spec
ldoktor: I'm not strongly against, I'm just pointing out that it's not everywhere (RHEL) and Fedora is not the only RPM-based distro. To me it'd make sense to simply list it. But again, not a strong opinion here.

PR #29, opened 2018-04-18
Encoding selection: use Python's locale module for choosing it

clebergnu:
The current encoding selection makes reference to an environment
variable that has no meaning outside of aexpect itself, and it's not
documented to serve that purpose.  In fact, it was a misinterpretation
of a similarly named environment variable, PYTHONIOENCODING.

Let's use the Python's standard library `locale` for choosing the
default encoding.  Since that is a guess and may return None, let's
set "UTF-8" as a fallback.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -14,16 +14,17 @@ API used to run/control interactive processes.
 
 :copyright: 2008-2015 Red Hat Inc.
 """
-import time
-import signal
-import os
-import sys
-import re
-import threading
-import shutil
-import select
+import locale
 import logging
+import os
+import re
+import select
+import shutil
+import signal
 import subprocess
+import sys
+import threading
+import time
 
 from aexpect.exceptions import ExpectError
 from aexpect.exceptions import ExpectProcessTerminatedError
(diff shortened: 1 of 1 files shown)


aexpect/client.py
```diff
@@ -123,9 +124,9 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
         self.a_id = a_id or data_factory.generate_random_string(8)
         self.log_file = None
         self.closed = False
-        # Use PYTHONENCODING or utf-8 (instead of ascii)
-        self.encoding = os.environ.get("PYTHONENCODING", "utf-8")
-
+        self.encoding = locale.getpreferredencoding()
```
ldoktor: Although I like the use of undeniably better `lang.getpreferredencoding()` I don't like the fact that this code still attempts to be smarter than python developers. I tried to break `getpreferredencoding` it resulted in ASCII and in sources I saw they do bunch of checks resulting usually in returning `sys.flags.utf8_mode` (which is not in my fedora-based python). Would you have anything against using it without the custom override?

Although I think it'd be useful to being able to force it on `__init__` as aexpect can be used on remote machines. Theoretically they can be running it on ASCII-only system, logging-in a utf-8 system which can immediately break the drainers...

ldoktor: I mean the scenario is:

```bash
LC_ALL=C python -c 'import aexpect; aexpect.ShellSession("socat /dev/ttyS1 -")'
```

PR #28, opened 2018-04-18
Tests: add a basic client/Spawn test

clebergnu:
And the very minimum test infrastructure.  This should allow one
to run the tests by doing the classical:

 $ python setup.py test

And of course can and will be extended/enabled on `make check`,
Travis-CI, etc.

Signed-off-by: Cleber Rosa <crosa@redhat.com>

PR #32, opened 2018-05-21
To support fd passing

luckyh:
aexpect.client: To support fd passing
    
Add new argument `pass_fds` for `Spawn`/`Tail`/`Expect`/`ShellSession`
and their related wrappers to support fd passing.
    
In python 2, the above utilities do support (implicit) fd passing, but
these would not work in python 3 for some reasons. By applying this
patch, now users can pass fds to the target process explicitly, and
implicit fd passing is not allowed.
    
### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -23,7 +23,6 @@ import threading
 import shutil
 import select
 import logging
-import subprocess
 
 from aexpect.exceptions import ExpectError
 from aexpect.exceptions import ExpectProcessTerminatedError
@@ -104,7 +103,7 @@ class Spawn(object):
     """
 
     def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
-                 linesep="\n"):
+                 linesep="\n", pass_fds=()):
         """
         Initialize the class and run command as a child process.
 
@@ -119,6 +118,8 @@ class Spawn(object):
                 parameter has an effect only when starting a new server.
         :param linesep: Line separator to be appended to strings sent to the
                 child process by sendline().
+        :param pass_fds: Optional sequence of file descriptors to keep open
+                between the parent and child.
         """
         self.a_id = a_id or data_factory.generate_random_string(8)
         self.log_file = None
@@ -176,11 +177,12 @@ class Spawn(object):
                 helper_cmd = utils_path.find_command(helper_versioned)
             except utils_path.CmdNotFoundError:
                 helper_cmd = utils_path.find_command(helper_noversion)
-            sub = subprocess.Popen([helper_cmd],
-                                   shell=True,
-                                   stdin=subprocess.PIPE,
-                                   stdout=subprocess.PIPE,
-                                   stderr=subprocess.STDOUT)
+            sub = utils_process.Popen([helper_cmd],
+                                      shell=True,
+                                      stdin=utils_process.PIPE,
+                                      stdout=utils_process.PIPE,
+                                      stderr=utils_process.STDOUT,
+                                      pass_fds=pass_fds)
             # Send parameters to the server
             sub.stdin.write(("%s\n" % self.a_id).encode(self.encoding))
             sub.stdin.write(("%s\n" % echo).encode(self.encoding))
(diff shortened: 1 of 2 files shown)


luckyh:
test_aexpect.py:
```python
import aexpect

f = open("/dev/null", "r")
p = aexpect.run_tail("./show_fd.py")
print(p.get_output())
p = aexpect.run_tail("./show_fd.py", pass_fds=[f.fileno()])
print(p.get_output())
```

show_fd.py:
```python
#!/usr/bin/python

import os
os.system("ls -l /proc/%s/fd" % os.getpid())
```

Test results:

python 2:
```
python2 test_aexpect.py 
total 0
lrwx------. 1 root root 64 May 21 15:45 0 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:45 1 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:45 2 -> /dev/pts/1
l-wx------. 1 root root 64 May 21 15:45 3 -> /tmp/aexpect_WUOQUOv6/server-log

total 0
lrwx------. 1 root root 64 May 21 15:45 0 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:45 1 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:45 2 -> /dev/pts/1
lr-x------. 1 root root 64 May 21 15:45 3 -> /dev/null
lrwx------. 1 root root 64 May 21 15:45 4 -> /tmp/aexpect_mwpt6WsS/lock-client-starting
lr-x------. 1 root root 64 May 21 15:45 5 -> /tmp/aexpect_WUOQUOv6/outpipe-tail
l-wx------. 1 root root 64 May 21 15:45 6 -> /tmp/aexpect_mwpt6WsS/server-log

```

python 3:
```
$ python3 test_aexpect.py 
total 0
lrwx------. 1 root root 64 May 21 15:42 0 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:42 1 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:42 2 -> /dev/pts/1

total 0
lrwx------. 1 root root 64 May 21 15:42 0 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:42 1 -> /dev/pts/1
lrwx------. 1 root root 64 May 21 15:42 2 -> /dev/pts/1
lr-x------. 1 root root 64 May 21 15:42 3 -> /dev/null

```

aexpect/utils/process.py
```diff
@@ -102,3 +104,32 @@ def process_in_ptree_is_defunct(ppid):
...
             break
     return defunct
+
+
+PIPE = subprocess.PIPE
+STDOUT = subprocess.STDOUT
+
+
+if sys.version_info < (3, 2):
+    # Actually we should rewrite `subprocess.Popen` with a complete
```
ldoktor: Well actually we can just use `subprocess32` backport and set the `fcntl` there (either as suggested here, or by using `subprocess._set_cloexec(fd, False)`, but as that one is private I'd suggest the `fcntl`.

aexpect/client.py
```diff
@@ -1378,12 +1396,15 @@ def run_fg(command, output_func=None, output_prefix="", timeout=1.0):
...
+    :param pass_fds: Optional sequence of file descriptors to keep open
+            between the parent and child.
 
     :return: A 2-tuple containing the exit status of the process and its
             STDOUT/STDERR output.  If timeout expires before the process
             terminates, the returned status is None.
     """
-    bg_process = run_bg(command, None, output_func, output_prefix, timeout)
+    bg_process = run_bg(command, None, output_func, output_prefix, timeout,
+                        pass_fds=pass_fds)
```
ldoktor: yep, this commit looks good.

ldoktor:
I see so without this change on py2 even the first process gets the `/dev/null` fd and on py3 no process get's the `/dev/null`, while with those fixes first execution masks the `/dev/null` and second passes it to both py2/py3.

Dangerous thing is that some people might depend on the old behavior so I'm not sure we can require the `pass_fds` on py2. On the other hand the usage of such feature is quite limited so perhaps we can try to educate people and require that, unless it breaks someones setup too badly.

So I like this approach, the only thing I'd change is the use of `subprocess32` on py2 instead of custom behavior (with the `fcntl` to mask the `FD_CLOEXEC`) as that one is a standard interface and could save us some troubles.

It'd be also nice if you could add the selftest along with check that `/dev/null` is handled properly. It doesn't have to be fancy, just use `in` or `not in` the output. There are currently no tests, but in https://github.com/avocado-framework/aexpect/pull/28 they are about to be introduced. You can simply create this test in `tests/test_pass_fds.py` and not care about the execution for now.

luckyh:
@ldoktor created https://github.com/avocado-framework/aexpect/pull/33 per your suggestion, thanks.

PR #33, opened 2018-05-22
To support fd passing [v2]

luckyh:
aexpect.client: To support fd passing

Add new argument `pass_fds` for `Spawn`/`Tail`/`Expect`/`ShellSession`
and their related wrappers to support fd passing.

In python 2, the above utilities do support (implicit) fd passing, but
these would not work in python 3 for some reasons. By applying this
patch, now users can pass fds to the target process explicitly, and
implicit fd passing is not allowed.

Signed-off-by: Xu Han xuhan@redhat.com

----
Changes:
```
v2:  replace subprocess with subprocess32
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ all:
 	@echo "make clean - Get rid of scratch and byte files"
 	@echo "make source - Create source package"
 	@echo "make install - Install on local system"
+	@echo "make requirements - Install runtime requirements"
 	@echo "make build-deb-src - Generate a source debian package"
 	@echo "make build-deb-bin - Generate a binary debian package"
 	@echo "make build-deb-all - Generate both source and binary debian packages"
@@ -35,6 +36,13 @@ source-release: clean
 
 install:
 	$(PYTHON) setup.py install --root $(DESTDIR) $(COMPILE)
+
+pip:
+	$(PYTHON) -m pip --version || $(PYTHON) -c "import os; import sys; import urllib; f = urllib.urlretrieve('https://bootstrap.pypa.io/get-pip.py')[0]; os.system('%s %s' % (sys.executable, f))"
+
+requirements: pip
+	- pip install "pip>=6.0.1"
+	- pip install -r requirements.txt
 
 prepare-source:
 	# build the source package in the parent directory
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -23,7 +23,7 @@ import threading
 import shutil
 import select
 import logging
-import subprocess
+import subprocess32 as subprocess
 
 from aexpect.exceptions import ExpectError
 from aexpect.exceptions import ExpectProcessTerminatedError
@@ -104,7 +104,7 @@ class Spawn(object):
     """
 
     def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
-                 linesep="\n"):
+                 linesep="\n", pass_fds=()):
         """
         Initialize the class and run command as a child process.
 
@@ -119,6 +119,8 @@ class Spawn(object):
                 parameter has an effect only when starting a new server.
         :param linesep: Line separator to be appended to strings sent to the
                 child process by sendline().
+        :param pass_fds: Optional sequence of file descriptors to keep open
+                between the parent and child.
         """
         self.a_id = a_id or data_factory.generate_random_string(8)
         self.log_file = None
@@ -180,7 +182,8 @@ class Spawn(object):
                                    shell=True,
                                    stdin=subprocess.PIPE,
                                    stdout=subprocess.PIPE,
-                                   stderr=subprocess.STDOUT)
+                                   stderr=subprocess.STDOUT,
+                                   pass_fds=pass_fds)
             # Send parameters to the server
             sub.stdin.write(("%s\n" % self.a_id).encode(self.encoding))
             sub.stdin.write(("%s\n" % echo).encode(self.encoding))
@@ -452,7 +455,7 @@ class Tail(Spawn):
     def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
                  linesep="\n", termination_func=None, termination_params=(),
                  output_func=None, output_params=(), output_prefix="",
-                 thread_name=None):
+                 thread_name=None, pass_fds=()):
         """
         Initialize the class and run command as a child process.
 
@@ -479,6 +482,8 @@ class Tail(Spawn):
                 output line.
         :param output_prefix: String to prepend to lines sent to output_func.
         :param thread_name: Name of thread to better identify hanging threads.
+        :param pass_fds: Optional sequence of file descriptors to keep open
+                between the parent and child.
         """
         # Add a reader and a close hook
         self._add_reader("tail")
@@ -486,7 +491,8 @@ class Tail(Spawn):
         self._add_close_hook(Tail._close_log_file)
 
         # Init the superclass
-        Spawn.__init__(self, command, a_id, auto_close, echo, linesep)
+        Spawn.__init__(self, command, a_id, auto_close, echo, linesep,
+                       pass_fds)
         if thread_name is None:
             self.thread_name = "tail_thread_%s_%s" % (self.a_id,
                                                       str(command)[:10])
@@ -669,7 +675,7 @@ class Expect(Tail):
     def __init__(self, command=None, a_id=None, auto_close=True, echo=False,
                  linesep="\n", termination_func=None, termination_params=(),
                  output_func=None, output_params=(), output_prefix="",
-                 thread_name=None):
+                 thread_name=None, pass_fds=()):
         """
         Initialize the class and run command as a child process.
 
@@ -695,6 +701,8 @@ class Expect(Tail):
         :param output_params: Parameters to send to output_func before the
                 output line.
         :param output_prefix: String to prepend to lines sent to output_func.
+        :param pass_fds: Optional sequence of file descriptors to keep open
+                between the parent and child.
         """
         # Add a reader
         self._add_reader("expect")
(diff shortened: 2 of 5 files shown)


luckyh:
```
$ python2 -m unittest test_pass_fds
....
----------------------------------------------------------------------
Ran 4 tests in 8.979s

OK
```
```
$ python3 -m unittest test_pass_fds
/usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9764 is still running
  ResourceWarning, source=self)
/usr/lib64/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedWriter name=6>
  _cleanup()
/usr/lib64/python3.6/subprocess.py:596: ResourceWarning: unclosed file <_io.BufferedReader name=7>
  _cleanup()
/usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9768 is still running
  ResourceWarning, source=self)
./usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9774 is still running
  ResourceWarning, source=self)
/usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9778 is still running
  ResourceWarning, source=self)
./usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9783 is still running
  ResourceWarning, source=self)
/usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9787 is still running
  ResourceWarning, source=self)
./usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9791 is still running
  ResourceWarning, source=self)
/usr/lib64/python3.6/subprocess.py:766: ResourceWarning: subprocess 9795 is still running
  ResourceWarning, source=self)
.
----------------------------------------------------------------------
Ran 4 tests in 9.262s

OK
sys:1: ResourceWarning: unclosed file <_io.BufferedWriter name=6>
sys:1: ResourceWarning: unclosed file <_io.BufferedReader name=7>
```

luckyh:
@ldoktor seems we need to open another pr to add `subprocess32` as a new dependence for packing the next rpm package.

aexpect/client.py
```diff
@@ -23,7 +23,7 @@
 import shutil
 import select
 import logging
-import subprocess
+import subprocess32 as subprocess
```
ldoktor: Please don't use backport as default. Newer python should use the stock subprocess. You can use something like:

```python
if sys.version_info[0:2] < (3, 2):
    import subprocess32 as subprocess
else:
    import subprocess
```

Similarly should be treated the `setup.py` install_requires (which are currently none) and the RPM specfile.

All changes have to be in a single commit as they are atomic (you need the subprocess32 and without this change you don't need subprocess32)

Makefile
```diff
@@ -36,6 +37,13 @@ source-release: clean
 install:
 	$(PYTHON) setup.py install --root $(DESTDIR) $(COMPILE)
 
+pip:
+	$(PYTHON) -m pip --version || $(PYTHON) -c "import os; import sys; import urllib; f = urllib.urlretrieve('https://bootstrap.pypa.io/get-pip.py')[0]; os.system('%s %s' % (sys.executable, f))"
+
+requirements: pip
+	- pip install "pip>=6.0.1"
+	- pip install -r requirements.txt
```
ldoktor: I'd suggest putting the `subprocess32` directly into `setup.py` as `install_requires` keyword and only if it's necessary (based on the suggested condition)

Makefile
```diff
@@ -36,6 +37,13 @@ source-release: clean
 install:
 	$(PYTHON) setup.py install --root $(DESTDIR) $(COMPILE)
 
+pip:
+	$(PYTHON) -m pip --version || $(PYTHON) -c "import os; import sys; import urllib; f = urllib.urlretrieve('https://bootstrap.pypa.io/get-pip.py')[0]; os.system('%s %s' % (sys.executable, f))"
```
ldoktor: Please don't (ever) try installing pip without user intervention. That's really intrusive. In case pip is not available the user will see the error. Alternatively you can simply suggest some recommendation by "echo" but not downloading it automatically as users should always favor their distribution channels to `get-pip` script.

tests/test_pass_fds.py
```diff
@@ -0,0 +1,96 @@
...
+
+import time
+import unittest
+
+from aexpect import client
+
+
+SLEEP = 1
+
+DEVNULL = "/dev/null"
```
ldoktor: more portable way is to use `os.devnull` but as the next command is posix only, we can hardcode it (perhaps it'd make sense to add `@skipUnless(os.path.exists('/proc/1/fd'), 'This test only works on posix-like systems')`.

tests/test_pass_fds.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def test_pass_fds_spawn(self):
+        """
+        Tests fd passing for `client.Spawn`
+        """
+        with open(DEVNULL, "r") as devnull:
+            fd = devnull.fileno()
+
+            child = client.Spawn(LIST_FD_CMD)
+            time.sleep(SLEEP)
```
ldoktor: Hmm, I really hate sleeps. How about `child.get_status()` instead? That will makes certain the command finished (and you can also check the status to avoid false-positives).

tests/test_pass_fds.py
```diff
@@ -0,0 +1,96 @@
...
+            self.assertFalse(DEVNULL in child.get_output())
+            child.close()
+
+            child = client.Spawn(LIST_FD_CMD, pass_fds=[fd])
+            time.sleep(SLEEP)
+            self.assertTrue(DEVNULL in child.get_output())
+            child.close()
+
+    def test_pass_fds_tail(self):
```
ldoktor: Hmm I don't think it's necessary to include the same test for each class which are simply inherited. Yes, we might accidentally remove it from some of them, but that is really improbable. If you want to be super-cautious just test `Spawn` and `ShellSession`, but I see no reason doing so.

tests/test_pass_fds.py
```diff
@@ -0,0 +1,96 @@
...
+
+            child = client.ShellSession(LIST_FD_CMD)
+            time.sleep(SLEEP)
+            self.assertFalse(DEVNULL in child.get_output())
+            child.close()
+
+            child = client.ShellSession(LIST_FD_CMD, pass_fds=[fd])
+            time.sleep(SLEEP)
+            self.assertTrue(DEVNULL in child.get_output())
+            child.close()
```
ldoktor: Please include the

```python
if __name__ == '__main__':
    unittest.main()
```

tests/test_pass_fds.py
luckyh: ok, I will only keep the `Spawn` one.

tests/test_pass_fds.py
luckyh: ok.

tests/test_pass_fds.py
luckyh: ok, makes sense, let me check it.

tests/test_pass_fds.py
luckyh: ok, let me check it.

Makefile
luckyh: ok.

Makefile
luckyh: I just copied these lines from avocado's, so, we would not need this anymore according to the above comment.

aexpect/client.py
luckyh: I have checked that the `python-subprocess32` rpm package is in fedora repository, but it is neither in rhel repository nor in epel repository, so we may have the problem.

ldoktor: You can ignore the RPM change for now, I can send it afterwards, but the missing `python-subprocess32` in epel looks more serious :-(. @clebergnu any ideas? Do you know how hard is it to add it there? And yes, we do need to add it to SPEC file even when it's in `setup.py`.

ldoktor: The only (sane) way to solve this that I can think of is to attempt to get `subprocess32` on older python and in case it fails use your simplified backport. In SPEC-file I can make the `subprocess32` dependency conditional based on RHEL/Fedora.

That way on py3.2+ we'll get the up-to-date subprocess, on py3.1- we'll get subprocess32 if available and custom backport in case it's not available (RHEL from RPM)

luckyh: I see the package is ready for new rhel releases..

luckyh: btw, I see the following content in `subprocess32.python3_redirect.__init__`
```python
"""When installed as subprocess32, this sets up a redirect to subprocess."""

import subprocess
import sys
if sys.version_info[:2] < (3,3):
    raise ImportError('Ancient Python 3 versions are not supported.')
# Doing this could crash some older Python interpreters due to the module
# reference going away before the import is complete?
sys.modules['subprocess32'] = subprocess
```
So that means it does not even support py3.1- either.

Makefile
clebergnu: @ldoktor how is installing pip different from the other requirements?

luckyh:
v3: https://github.com/avocado-framework/aexpect/pull/34

aexpect/client.py
ldoktor: Well in Avocado we support py3.4, we can do the same in aexpect and simply use `if sys.version_info[0] < 3`.

Makefile
ldoktor: The difference is that dependencies are installed via `pip` which can be set accordingly, while installing pip from external source automatically seems quite intrusive to me. When people have it already and ask `make requirements` to get requirements, sure. No problem. But when they don't have pip at all I think we should only inform the user about possible means to get it.

clebergnu: @ldoktor so it **seems** intrusive to you, and you **think** it shouldn't be done.  But in reality, both will download/install software when a user run `make requirements` and doesn't have them installed.

ldoktor: ... anyway I think we should move the discussion to Avocado as it's not related to this PR (anymore)

PR #40, opened 2018-06-08
Tests: add a basic client/Spawn test [v2]

clebergnu:
And the very minimum test infrastructure.  This should allow one
to run the tests by doing the classical:

 $ python setup.py test

And of course can and will be extended/enabled on `make check`,
Travis-CI, etc.

---

Changes from v1 (#28):
 * The test itself was completely rewritten (and simplified)

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -181,8 +181,9 @@ class Spawn(object):
             # try to find python specific version of aexpect-helper first, then
             # try unversioned
             helper_noversion = 'aexpect-helper'
-            helper_versioned = '{0}-{1.major}.{1.minor}'.format(
-                helper_noversion, sys.version_info)
+            helper_versioned = '{0}-{1}.{2}'.format(helper_noversion,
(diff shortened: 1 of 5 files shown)


aexpect/client.py
```diff
@@ -181,8 +181,9 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
             # try to find python specific version of aexpect-helper first, then
             # try unversioned
             helper_noversion = 'aexpect-helper'
-            helper_versioned = '{0}-{1.major}.{1.minor}'.format(
-                helper_noversion, sys.version_info)
+            helper_versioned = '{0}-{1}.{2}'.format(helper_noversion,
+                                                    sys.version_info[0],
+                                                    sys.version_info[1])
```
ldoktor: Interesting to remove py2.6 testing and then to introduce 2.6 compat fix. Anyway I do prefer 2.6 compatibility (even when not official) so yes to this.

PR #42, opened 2018-07-03
client: Fix shell status error

luckyh:
Some type of shell (e.g. windows batch) would return a negative errno
on command failure, we should handle this, otherwise, would get a
`ShellStatusError`.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -976,6 +976,9 @@ class ShellSession(Expect):
     process for responsiveness.
     """
 
+    # Return code pattern of shell interpreter
+    __RE_STATUS = re.compile("^-?[0-9]+$")
+
     def __init__(self, command=None, a_id=None, auto_close=True, echo=False,
                  linesep="\n", termination_func=None, termination_params=(),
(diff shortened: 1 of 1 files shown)


aexpect/client.py
```diff
@@ -1240,7 +1240,8 @@ def cmd_status_output(self, cmd, timeout=60, internal_timeout=None,
             raise ShellStatusError(cmd, out)
 
         # Get the first line consisting of digits only
-        digit_lines = [l for l in status.splitlines() if l.strip().isdigit()]
+        digit_lines = [l for l in status.splitlines()
+                       if re.match("^-?[0-9]+$", l.strip())]
```
ldoktor: good catch. Would you please compile the regexp and store it as `self.RE_STATUS` so it doesn't have to be compiled per each line.

ldoktor: ... or rather `self.__RE_STATUS`

luckyh: Sure, thanks.

luckyh:
@ldoktor have updated per your suggestion, thanks.

PR #43, opened 2018-07-05
client: Fix the issue of getting command status

luckyh:
When a system is under heavy load, calling `s.cmd_status_output()`
might get `ShellStatusError`, the problem is the command of
getting status may not finish in 10s (the current value).

The patch tries to solve the problem by enlarging the value to 30s.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1237,7 +1237,7 @@ class ShellSession(Expect):
         out = self.cmd_output(cmd, timeout, internal_timeout, print_func, safe)
         try:
             # Send the 'echo $?' (or equivalent) command to get the exit status
-            status = self.cmd_output(self.status_test_command, 10,
+            status = self.cmd_output(self.status_test_command, 30,
                                      internal_timeout, print_func, safe)
         except ShellError:
             raise ShellStatusError(cmd, out)


ldoktor:
Actually I can imagine one example where the status command might depend on previous activity and your solution would possibly help. Copying a big file with great timeout might result in flush between command and the status command. Anyway we need to draw a line and here I think having predictable timeout is better (imagine people executing long-lasting commands that failed and hanged. They'd have to wait for hours/days for the status command to timeout.

luckyh:
Well, I can understand your concern that it is better to set timeout with a max limit, however it is hard to get a certain value for it since the duration of the status command would be quite long under some extreme conditions. To support these conditions we can make it user-configurable, but I do not want the interfaces being so complicated.

In my case (sending `echo %errorlevel%` to a windows vm with heavy load), the duration was around 15s, so it is okay for me to enlarge the value to 30s to solve the problem.

ldoktor:
Btw just for the record the perfect commit message says what is changed instead of just generic saying... anyway let's not be pedantic and merge it :smile:

luckyh:
@ldoktor @lmr shall we have a new minor release to contain this and https://github.com/avocado-framework/aexpect/pull/42 ? thanks!

ldoktor:
Well there are just 2 minor fixes + bunch of style fixes... Not sure whether it's worth it: https://github.com/avocado-framework/aexpect/compare/1.5.1...master so depends on you, @lmr...

PR #45, opened 2018-07-20
Use select.poll for waiting on file descriptors

### LINKED ISSUE #44: RFC: Use select.poll for waiting on file descriptors
I'm running stress_boot.py [1] from the avocado-vt plugin and avocado is failing to boot the 60th (or so) guest:
```
L0651 DEBUG| [stderr] Traceback (most recent call last):
L0651 DEBUG| [stderr]   File "/usr/lib64/python2.7/threading.py", line 812, in __bootstrap_inner
L0651 DEBUG| [stderr]     self.run()
L0651 DEBUG| [stderr]   File "/usr/lib64/python2.7/threading.py", line 765, in run
L0651 DEBUG| [stderr]     self.__target(*self.__args, **self.__kwargs)
L0651 DEBUG| [stderr]   File "/usr/lib/python2.7/site-packages/aexpect/client.py", line 584, in _tail
L0651 DEBUG| [stderr]     r, w, x = select.select([fd], [], [], 0.05)
L0651 DEBUG| [stderr] ValueError: filedescriptor out of range in select()
```

`ls -l /proc/<pid>/fd | wc -l` shows ~1040 by the time this fails.

From `man select` I see that this is expected:
> select() can monitor only file descriptors numbers that are less than FD_SETSIZE; poll(2) does not have this limitation.  See BUGS.

`select.select` is used in aexpect in 3 places, here is one of them:
```
                try:
                    # See if there's any data to read from the pipe
                    read_available_fds = select.select([tail_pipe], [], [],
                                                       0.05)[0]
                except (select.error, TypeError):
                    break
                if tail_pipe in read_available_fds:
                    # Some data is available; read it
                    new_data = os.read(tail_pipe, 1024)
```

So, I altered the code to use `select.poll` in aexpect (and virttest) and stress_boot can now boot 200 guests.

I'm still to rebase it on top of the latest version but I wanted to hear what you guys think first. I'm not sure about the compatibilty with other platforms, I see that pexpect has a switch to choose between select and poll: https://github.com/pexpect/pexpect/commit/6db8d9b8e91a49c75bf67a95f245733d357770a8 

---
More info:
```
$ ulimit -n
65536

$ uname -r
4.11.0-44.el7a.ppc64le

$ rpm -qa | grep "avocado\|aexpect"
python2-aexpect-1.5.1-1.el7.noarch
python2-avocado-62.0-0.el7.noarch
avocado-plugins-vt-62.0-0.el7.noarch
python-avocado-common-62.0-0.el7.noarch

$ avocado run io-github-autotest-qemu.stress_boot --vt-type qemu --vt-arch ppc64le --vt-extra-params nettype=user max_vms=200
```
1- https://github.com/autotest/tp-qemu/blob/master/generic/tests/stress_boot.py

fabianorosas:
This patchset replaces select() with poll() so that we can avoid "ValueError: filedescriptor out of range in select()" when spawning several aexpect processes.

Tested via avocado with:
```
avocado run io-github-autotest-qemu.qmp_command.qmp_query-status \
 --vt-type qemu --vt-arch ppc64le --vt-extra-params nettype=user
```
and

```
avocado run io-github-autotest-qemu.stress_boot \
 --vt-type qemu --vt-arch ppc64le --vt-extra-params nettype=user max_vms=250
```
The above is an improvement from the previous state where the test would fail with max_vms = 60

This also introduces a change in behavior where tail threads are no longer allowed to start if the `close()` method has already been called, i.e. the child process is not running anymore and the readers have already been removed.

Closes #44

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -525,8 +525,9 @@ class Tail(Spawn):
 
         # Start the thread in the background
         self.tail_thread = None
-        if termination_func or output_func:
-            self._start_thread()
+        if self.is_alive():
+            if termination_func or output_func:
+                self._start_thread()
 
     def __reduce__(self):
         return self.__class__, (self.__getinitargs__())
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -614,6 +614,8 @@ class Tail(Spawn):
 
         try:
             tail_pipe = self._get_fd("tail")
+            poller = select.poll()
+            poller.register(tail_pipe, select.POLLIN)
             bfr = ""
             while True:
                 if _THREAD_KILL_REQUESTED.is_set():
@@ -624,11 +626,10 @@ class Tail(Spawn):
                     return
                 try:
                     # See if there's any data to read from the pipe
-                    read_available_fds = select.select([tail_pipe], [], [],
-                                                       0.05)[0]
-                except (select.error, TypeError):
+                    poll_status = poller.poll(50)
+                except select.error:
                     break
-                if tail_pipe in read_available_fds:
+                if poll_status:
                     # Some data is available; read it
                     new_data = os.read(tail_pipe, 1024)
                     if not new_data:
@@ -749,19 +750,22 @@ class Expect(Tail):
         :param timeout: Timeout for reading child process output.
         """
         if internal_timeout is None:
-            internal_timeout = 0.1
+            internal_timeout = 100
+        else:
+            internal_timeout *= 1000
         end_time = None
         if timeout:
             end_time = time.time() + timeout
         expect_pipe = self._get_fd("expect")
+        poller = select.poll()
+        poller.register(expect_pipe, select.POLLIN)
         data = ""
         while True:
             try:
-                read_available_fds = select.select([expect_pipe], [], [],
-                                                   internal_timeout)[0]
-            except (select.error, TypeError):
+                poll_status = poller.poll(internal_timeout)
+            except select.error:
                 return data
-            if expect_pipe in read_available_fds:
+            if poll_status:
                 new_data = os.read(expect_pipe, 1024).decode(self.encoding,
                                                              "ignore")
                 if not new_data:
@@ -842,16 +846,17 @@ class Expect(Tail):
         if not match_func:
             match_func = self.match_patterns
         expect_pipe = self._get_fd("expect")
+        poller = select.poll()
+        poller.register(expect_pipe, select.POLLIN)
         output = ""
         end_time = time.time() + timeout
         while True:
             try:
-                _timeout = max(0, end_time - time.time())
-                read_available_fds = select.select([expect_pipe], [], [],
-                                                   _timeout)[0]
-            except (select.error, TypeError):
+                poll_timeout_ms = max(0, (end_time - time.time()) * 1000)
+                poll_status = poller.poll(poll_timeout_ms)
+            except select.error:
                 break
-            if not read_available_fds:
+            if not poll_status:
                 raise ExpectTimeoutError(patterns, output)
             # Read data from child
             data = self.read_nonblocking(internal_timeout,


fabianorosas:
Depends on: https://github.com/avocado-framework/aexpect/pull/46 so that the CI checks pass.

aexpect/client.py
```diff
@@ -525,8 +525,9 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
 
         # Start the thread in the background
         self.tail_thread = None
-        if termination_func or output_func:
-            self._start_thread()
+        if self.is_alive():
+            if termination_func or output_func:
+                self._start_thread()
```
ldoktor: Nice catch, even though the thread dies immediately it's better not to run it in the first place.

ldoktor: Would you please include a selftest?

ldoktor: Actually without modifying the sources it'd be hard to see whether the `_start_thread` had been called or not, so unless you have an idea how to cleanly do that let me just add card to add `pickle` tests...: https://trello.com/c/GODRyeHr/1385-aexpect-add-pickle-selftests

aexpect/client.py
```diff
@@ -624,11 +626,10 @@ def _print_line(text):
                     return
                 try:
                     # See if there's any data to read from the pipe
-                    read_available_fds = select.select([tail_pipe], [], [],
-                                                       0.05)[0]
-                except (select.error, TypeError):
+                    read_available_fd = poller.poll(50)
```
ldoktor: well `poller.poll` returns list of (fd, event), therefor `read_available_fd` is a misleading name. As we know we only registered tail_pipe, how about calling it `poll_status`?

aexpect/client.py
```diff
@@ -750,18 +751,20 @@ def read_nonblocking(self, internal_timeout=None, timeout=None):
         """
         if internal_timeout is None:
             internal_timeout = 0.1
+        poll_timeout_ms = internal_timeout * 1000
```
ldoktor: internal_timeout is only used for poll, how about:

```python
if internal_timeout is None:
    internal_timeout = 100
else:
    internal_timeout *= 1000
```

aexpect/client.py
```diff
@@ -750,18 +751,20 @@ def read_nonblocking(self, internal_timeout=None, timeout=None):
...
         expect_pipe = self._get_fd("expect")
+        poller = select.poll()
+        poller.register(expect_pipe, select.POLLIN)
         data = ""
         while True:
             try:
-                read_available_fds = select.select([expect_pipe], [], [],
-                                                   internal_timeout)[0]
-            except (select.error, TypeError):
+                read_available_fd = poller.poll(poll_timeout_ms)
```
ldoktor: Again the `read_available_fd` is not accurate.

aexpect/client.py
```diff
@@ -842,16 +845,17 @@ def read_until_output_matches(self, patterns, filter_func=lambda x: x,
...
         output = ""
         end_time = time.time() + timeout
         while True:
             try:
-                _timeout = max(0, end_time - time.time())
-                read_available_fds = select.select([expect_pipe], [], [],
-                                                   _timeout)[0]
-            except (select.error, TypeError):
+                poll_timeout_ms = max(0, (end_time - time.time()) * 1000)
+                read_available_fd = poller.poll(poll_timeout_ms)
```
ldoktor: yep, `poll_timeout_ms` is better name, but the `read_available_fd` not...

aexpect/client.py
fabianorosas: I'd need to take a closer look at the code to figure out `pickle` tests. Let's leave that on your side for now. If I come up with something before someone starts on that card I'll get in touch.

fabianorosas:
I'm working on the suggested changes.

PR #8, opened 2016-08-09
Support for python3

ldoktor:
This extends the initial python3 support from https://github.com/autotest/aexpect/pull/7. Not every functionality was tested on python3, but the basic workflow with `ShellSession` works well (using python2 aexpect-worker, which needs to be adjusted before declaring aexpect fully python3 compatible)

PR #48, opened 2018-09-17
Add missing match '[1S' for output

kylazhang:
May get output like '[1S' which need to match

Signed-off-by: Kylazhang <weizhan@redhat.com>

### DIFF
--- a/aexpect/utils/astring.py
+++ b/aexpect/utils/astring.py
@@ -36,7 +36,7 @@ def strip_console_codes(output, custom_codes=None):
     index = 0
     output = "\x1b[m%s" % output
     console_codes = "%[G@8]|\\[[@A-HJ-MPXa-hl-nqrsu\\`]"
-    console_codes += "|\\[[\\d;]+[HJKgqnrm]|#8|\\([B0UK]|\\)"
+    console_codes += "|\\[[\\d;]+[HJKSgqnrm\\d]?|#8|\\([B0UK]|\\)|O[P-S]"
     if custom_codes is not None and custom_codes not in console_codes:
         console_codes += "|%s" % custom_codes
     while index < len(output):


ldoktor:
Oh I see, there are things like:

```python
'[1;2S': 'S-F4'
'[1;3S': 'M-F4',
'[1;5S': 'C-F4'
'[1;6S': 'S-C-F4'
```

ldoktor:
Anyway if we add `S` we should also include the rest of the codes: http://wiki.bash-hackers.org/scripting/terminalcodes most of them are not covered. Maybe simplifying it ... What do you think, @kylazhang?

kylazhang:
@ldoktor I get this issues during my test and '[1S' can not be match so report error, I update the patch to allow match for all esc_codes, is that ok?

ldoktor:
And is it really just `[1S`? Because I don't know what that means. Do you have a sample? Maybe it's an error as I don't see it to be defined...

kylazhang:
@ldoktor I don't know what that means either, but I get it from the log, you can have a look, thanks
https://privatebin-it-iso.int.open.paas.redhat.com/?df52db94faa81c1e#K+m0GgSY+5uFIcQmVfh1dIq8JpzqEo1yGsUbKEppSXY=
and before '********' is the pure string, after is the corresponding codes

ldoktor:
Well these are no console codes, these are binary data being produced... @lmr @clebergnu we should decide which way to go. Do we want this function to strip any garbage, or do we want it to strip purely shell escape codes?

lmr:
@ldoktor My idea would be to include a function `sanitize_output()` for `aexpect.astring`, that removes data such as the one reported by @kylazhang and calls remove_console_codes, so that we have the clean output, and update the internal callers to use `sanitize_output()`. With this we'd break no expected behavior and we'd still have the desired effect.

ldoktor:
One thing I don't understand is why do you need to remove those characters? They are valid ascii characters and are part of the output (maybe can we get in touch via IRC/email to discuss this?)

clebergnu:
As @ldoktor points out, the output to be filtered (or sanitized) may be something else, and not a terminal control code. I think it makes for aexpect to accept a filter function as a parameter, and the user can define what to do with the raw data before the match.

kylazhang:
I call vm.serial_console.get_stripped_output() which will call 
```
    def get_stripped_output(self):
        """
        Return the STDOUT and STDERR output without the console codes escape
        and sequences of the process so far.
        """
        return astring.strip_console_codes(self.get_output())
```
in aexpect/client.py and can not apply for extra parameters here, so when call strip_console_codes in astring it will report error 
ValueError: [1S is not included in the known console codes list %[G@8]|\[[@A-HJ-MPXa-hl-nqrsu\`]|\[[\d;]+[HJKgqnrm]|#8|\([B0UK]|\)

Do you mean we need give extra parameter for get_stripped_output()?

ldoktor:
Well looking at it from a distance I think the main issue here is not the stripping, but the ValueError. The output can contain anything and we should either strip the console codes, or simply return the output including things that looked a bit like console codes, but are not console codes (as we know them).

There for I'd prefer returning the original output instead of "raise ValueError" in case of IndexError. Alternatively, to avoid breaking the expectations, we might want to do this optionally based on another argument to `strip_console_codes`. @clebergnu does that sounds good to you?

(note that would mean your workflow would not crash, but it'd return some weird characters, but that should IMO be expected as you are feeding it from binary bios outputs).

pevogam:
> Well looking at it from a distance I think the main issue here is not the stripping, but the ValueError. The output can contain anything and we should either strip the console codes, or simply return the output including things that looked a bit like console codes, but are not console codes (as we know them).
> 
> There for I'd prefer returning the original output instead of "raise ValueError" in case of IndexError. Alternatively, to avoid breaking the expectations, we might want to do this optionally based on another argument to `strip_console_codes`. @clebergnu does that sounds good to you?
> 
> (note that would mean your workflow would not crash, but it'd return some weird characters, but that should IMO be expected as you are feeding it from binary bios outputs).

I completely agree with @ldoktor about this. It is impossible to filter for all sorts invalid characters and we should not impose the assumption that this is a text output at all. I think the same issue of this restrictive assumptions keeps popping up in currently registered aexpect issues.

pevogam:
Given that a few years have passed since this issue and a different fix was already proposed and merged I would recommend closing this PR.

PR #52, opened 2019-04-02
support with statement

lolyu:
As the subject implies, this aims to enable with statement for 'Spawn'
and its descendant classes.

Signed-off-by: lolyu <lolyu@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -450,6 +450,12 @@ class Spawn(object):
         except OSError:
             pass
 
+    def __enter__(self):
+        return self
+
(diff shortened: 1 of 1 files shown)


aexpect/client.py
```diff
@@ -450,6 +450,13 @@ def send_ctrl(self, control_str=""):
         except OSError:
             pass
 
+    # for `with spawn(...) as session:`
```
hereischen: I think you can delete this comment

ldoktor:
Hello guys, sorry for such a late review, but yes, it makes sense and ought to be there for a long time...

Issue #50, opened 2019-02-07
Extending aexpect with file transfer


pevogam:
Is it possible to extend aexpect with simple data transfer functionality like the one in avocado-vt's [remote](https://github.com/avocado-framework/avocado-vt/blob/master/virttest/remote.py) module?

The reasoning behind this is that file transfer is tightly coupled (also in terms of use cases) to aexpect and generalizes beyond virtualization as the use of virtual machines done in avocado-vt.

ldoktor:
Actually I've been struggling to find the time to polish and refine a `SSHSession` class that inherits from `ShellSession` that allows reading/writing/... remote files and has other helpers with regards to ssh, but unfortunately I'm fairly overloaded now. If you have some spare cycles that is what I'd suggest as extending it of `scp` seems possible...

pevogam:
Will it be ok this is done by copying `remote.py` (and deps) from avocado-vt and then simply calling its file copy methods from a newly developed `SSHSession` instance?

ldoktor:
Aexpect should not depend on `avocado-vt` as `avocado-vt` already depends on `aexpect`. So at least for a while we'd have to maintain copy of the implementation in both and then slowly remove it from `avocado-vt` and use it from `aexpect` instead.

ldoktor:
Btw just to clarify I won't have time to propose the SSHSession any time soon due to other tasks so take that only as a recommendation, or suggestion, but feel free to propose a different way if it doesn't fit your goal.

pevogam:
@ldoktor Gladly, will be able to try to extend this and add a new utility about running code remotely in the near months.

pevogam:
Hi @ldoktor, I have implemented the minimal set of changes I saw fit in #55 where I didn't extend the previous client classes with an SSHSession since all remote operations (from `remote.py`) create and manipulate sessions and instances of these classes themselves.

Feel free to take a look when you can.

pevogam:
Since we now have such file transfer functionality and the additional session classes, I am closing this. Anybody with more specific file transfer feature request is free to open another issue.

PR #55, opened 2019-05-04
Remote utils extension

pevogam:
Add an extension to aexpect where we can log into remote machines and transfer files.

This functionality is available from avocado's remote utility.


ldoktor:
Hello @pevogam, this looks useful, but actually it just copies a feature from Avocado-vt to aexpect, which means we'd have to maintain 2 forks of the same feature. Better solution would be to create a `RemoteCommander` project under `avocado-framework` (I can set it up for you), generalize it (the way you did here) and then include it from Avocado-vt and Aexpect. The key here is not to depend on any of those projects to avoid circular dependency.

Note we have the same plan with the remaining avocado.utils. We want to extract, generalize and publish them independently. There just wasn't enough resources to do it yet. Doing it for remote-commander would be actually the first (well, counting aexpect it'd be the second) step to the right direction.

@clebergnu any ideas, suggestions?

pevogam:
> Hello @pevogam, this looks useful, but actually it just copies a feature from Avocado-vt to aexpect,

Precisely, just as we discussed in gh-50 where you already mentioned that "at least for a while we'd have to maintain copy of the implementation in both and then slowly remove it from `avocado-vt` and use it from `aexpect` instead".

> which means we'd have to maintain 2 forks of the same feature. Better solution would be to create a `RemoteCommander` project under `avocado-framework` (I can set it up for you), generalize it (the way you did here) and then include it from Avocado-vt and Aexpect. The key here is not to depend on any of those projects to avoid circular dependency.

I am not convinced this is a better option. Here some of my reasoning:

1) The `remote.py` utility is too coupled with `aexpect`, so much so that one could say it simply imports `aexpect` and adds a few new methods to it. I could say it uses aexpect sessions for all of its own functionality and thus is more like an extra module within aexpect than an entire project on its own.

2) Separating it in a separate project will introduce a large overhead of packaging and deployment and as you said resources are already scarce.

3) I have actually removed the `remote_commander` functionality which requires many extra migrated modules as well as some remote classes emulating `avocado.utils.process.run` call through a session command call. This means I am not even sure why we should namely start a remote commander project.

4) You already agreed that it is better to move the `remote.py` functionality to aexpect gradually so we are not striving for maintaining multiple copies at two repos for the long term. In fact, I might even assist in fixing the calls in `avocado-vt` to directly use `aexpect` and we might skip the entire duplication part altogether.

> 
> Note we have the same plan with the remaining avocado.utils. We want to extract, generalize and publish them independently. There just wasn't enough resources to do it yet. Doing it for remote-commander would be actually the first (well, counting aexpect it'd be the second) step to the right direction.
> 

Do you have some source of documentation for this remote commander? Is there a reason why it cannot be part of aexpect and the two are different enough to be separate projects? The way I see it so far is that `aexpect` is a way to manage and communicate with subprocesses and remote processes and all these parts fall within this scope. I am even planning to add a PR for remote code execution that so far I assigned to the same tool set.

pevogam:
The travis CI has all sorts of syntax linting problems almost none of which are related to this PR and most of which are already present in `avocado-vt` from where the larger portion of the code was taken. Any advice on how to proceed with this is highly appreciated but I guess we should complete the previous discussion first.

ldoktor:
Hello @pevogam, moving (not forking) it to aexpect works for me as well but I'd like to see some commitment to do the other part (avocado-vt) as well. No need to send PR I just want your promise to do it once this is ready (first attempt to import commander from aexpect and fallback to the original avocado-vt implementation which would be marked as deprecated and removed after several releases).

Now for remote commander I'm not the real user of this feature but @rdkdd is. Radek, would you please have some time to try it? I'll do my best but it'd be nice to have someone who's skilled with it approval (at least functional, no need to do thorough review).

As for the travis issues, aexpect checkers are way more strict than avocado ones that are still more strict than avocado-vt. On the other hand we don't require all commits to pass so it's fine to add last commit that will address all the style issues. I can help you with that after we hammer-out the details.

I'll try to do a thorough review in the following 1-2 weeks, but I can not promise anything as my primary assignment moved to other projects. Still aexpect is project I'm really fond of so I'll push hard to assign some time on this as I see you spent a lot of time on it already and I don't like wasting resources. Please bear with me (alternatively, @lmr and @clebergnu, review help is welcome).

pevogam:
> Hello @pevogam, moving (not forking) it to aexpect works for me as well but I'd like to see some commitment to do the other part (avocado-vt) as well. No need to send PR I just want your promise to do it once this is ready (first attempt to import commander from aexpect and fallback to the original avocado-vt implementation which would be marked as deprecated and removed after several releases).

I could invest some time in both repos in parallel, this is no problem. I didn't know what was the exact plan of action (and planned time window) for the deprecation on the avocado-vt side before but now it is clear.

> As for the travis issues, aexpect checkers are way more strict than avocado ones that are still more strict than avocado-vt. On the other hand we don't require all commits to pass so it's fine to add last commit that will address all the style issues. I can help you with that after we hammer-out the details.

Great, will ignore it for now until I have a running version that fulfills all our user cases without error, then we can take a look at the CI details.

> I'll try to do a thorough review in the following 1-2 weeks, but I can not promise anything as my primary assignment moved to other projects. Still aexpect is project I'm really fond of so I'll push hard to assign some time on this as I see you spent a lot of time on it already and I don't like wasting resources. Please bear with me (alternatively, @lmr and @clebergnu, review help is welcome).

Thanks for the sparing of resources, will move on but I just need a few more comments regarding the two possibilities of:

1) Implement `remote.py` and other modules managing session objects the way we currently did.
2) Do as I added in the EDIT here:

> _EDIT:_ There is also the possibility of turning the `remote_login` method into a constructor for an `SSHSession` instance of an extended class containing attributes like `address`, `port`, `username`, and `password`. We could then turn the `remote` utility into pure file transfer one which creates SSH or other sessions if a file transfer is requested from it.

Any preference and pros-cons I did not foresee?

rdkdd:
Hi @pevogam and @ldoktor ,
I think to remove this bit to `aexpect` is a good idea. But I miss `remote_commander` function in this new `remote.py` file, which @pevogam removed together with modules in `remote_commander` folder. I am using this functionality in my tests (see https://github.com/spiceqa/tp-spice/blob/530bad4f405479a6c3e9a477fbf09a99de91ce57/spice/tests/rv_gui.py#L140 and further) to run python script from within VM (which was copied there before) and load test parameters there (see also https://github.com/spiceqa/tp-spice/blob/master/spice/client-tests/simple.py). Actually `remote_commander` function is incorporated also in `virt_vm.py` file (#L1135). For that reason I can not verify functionality as @ldoktor asked me. 
To be honest code in whole `remote_commander` dir seems very complex to me and I do not understand much there. I think there should be an easier way to start python script from within VM with `params` defined in testcase. What you think?

pevogam:
Hi @rdkdd ,

> Hi @pevogam and @ldoktor ,
> I think to remove this bit to `aexpect` is a good idea. But I miss `remote_commander` function in this new `remote.py` file, which @pevogam removed together with modules in `remote_commander` folder. 

Sorry, I thought it is part of the same code and purpose. Can restore it easily.

> I am using this functionality in my tests (see https://github.com/spiceqa/tp-spice/blob/530bad4f405479a6c3e9a477fbf09a99de91ce57/spice/tests/rv_gui.py#L140 and further) to run python script from within VM (which was copied there before) and load test parameters there (see also https://github.com/spiceqa/tp-spice/blob/master/spice/client-tests/simple.py). Actually `remote_commander` function is incorporated also in `virt_vm.py` file (#L1135). For that reason I can not verify functionality as @ldoktor asked me.

So which version of the function are you using and we should keep? How about keeping the `remote_commander` at the `virt_vm.py` and keeping the current code migration focused on the pure `remote.py` file transfer and session management?

> To be honest code in whole `remote_commander` dir seems very complex to me and I do not understand much there. I think there should be an easier way to start python script from within VM with `params` defined in testcase. What you think?

This is one reason I don't feel like touching this code and including it in migration - I don't use it, I am not sure who does, and would prefer to change as little as possible for others and thus minimize the chances of breakage.

pevogam:
@ldoktor First draft of the migration on the avocado-vt side can be found [here](https://github.com/avocado-framework/avocado-vt/pull/2095). I have preserved all non-migrated code in an `remote_old` module here that could be deprecated later on if people like @rdkdd really don't use or need that code.

As I mentioned there, this will require a quite significant testing and I am not even sure how a possible simultaneous merge on both repos should be managed so feel free to guide me through any further reviews here and there.

pevogam:
> As I mentioned there, this will require a quite significant testing and I am not even sure how a possible simultaneous merge on both repos should be managed so feel free to guide me through any further reviews here and there.

Actually, I think it is shouldn't be too hard - make this merge possible first and once done (possibly with a release), look at the PR on the avocado-vt side.

aexpect/remote.py
```diff
@@ -826,6 +826,68 @@ def stop_server(session):
         d_session.close()
 
 
+def login_reusing_session(session, log_filename=None, log_function=None,
```
ldoktor: This is a bit misleading name. At first glance I thought you reuse the `session` connection, but with scp it became clear that is not the case. How about `login_based_on_session` or `login_from_session` (`scp_from|to_session`)?

aexpect/remote.py
```diff
@@ -0,0 +1,1297 @@
+"""
```
ldoktor: The license has to be here even though it was not part of the original source and there has to be a note about the original source like we do in avocado, for example here: https://github.com/avocado-framework/avocado/blob/master/avocado/utils/linux_modules.py

aexpect/remote.py
```diff
@@ -0,0 +1,1297 @@
...
+                continue
+            elif match == 8:  # "Warning added RSA"
+                if debug:
+                    logging.debug("Got 'Warning added RSA to known host list")
+                continue
+            elif match == 12:  # prompt
+                if debug:
+                    logging.debug("Got shell prompt -- logged in")
+                break
+            elif match == 13:  # console prompt
```
ldoktor: There were some new changes since then (just mentioning to not to forgat about that. No need to backport them yet)

aexpect/rss_client.py
```diff
@@ -0,0 +1,543 @@
+#!/usr/bin/python
```
ldoktor: The license here as well...

aexpect/remote.py
```diff
@@ -309,7 +317,7 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     if host and host.lower().startswith("fe80"):
         if not interface:
             raise LoginError("When using ipv6 linklocal an interface must "
-                             "be assigned")
+                             "be assigned", None)
```
ldoktor: Well the second argument is output, therefore '' would fit better here. Also I'd suggest making the argument optional (defaults to '') in the LoginError, rather than supplying empty output when raising the exc.

aexpect/remote.py
```diff
@@ -523,10 +527,10 @@ def scp_to_remote(host, port, username, password, local_path, remote_path,
...
     if directory:
         command = "%s -r" % command
-    command += (" -v -o UserKnownHostsFile=/dev/null "
-                "-o StrictHostKeyChecking=no "
-                "-o PreferredAuthentications=password %s "
-                "-P %s %s %s@\[%s\]:%s" %
+    command += (r" -v -o UserKnownHostsFile=/dev/null "
+                r"-o StrictHostKeyChecking=no "
+                r"-o PreferredAuthentications=password %s "
+                r"-P %s %s %s@\[%s\]:%s" %
```
ldoktor: well the `r` prefix is only required for the last line, but no big deal...

aexpect/remote.py
```diff
@@ -918,7 +922,7 @@ def transfer(*args, **kwargs):
 @throughput_transfer
 def copy_files_to(address, client, username, password, port, local_path,
                   remote_path, limit="", log_filename=None, log_function=None,
-                  verbose=False, timeout=600, interface=None, filesize=None,
+                  verbose=False, timeout=600, interface=None, _filesize=None,
```
ldoktor: Changing the argument names is tricky as people might be using them (even though they are not used later). Better solution is to either add `pylint: disable=...` or to use `del(filesize)` in the function body (it deletes it in the end anyway...)

aexpect/remote.py
```diff
@@ -961,7 +965,7 @@ def copy_files_to(address, client, username, password, port, local_path,
 @throughput_transfer
 def copy_files_from(address, client, username, password, port, remote_path,
                     local_path, limit="", log_filename=None, log_function=None,
-                    verbose=False, timeout=600, interface=None, filesize=None,
+                    verbose=False, timeout=600, interface=None, _filesize=None,
```
ldoktor: the same here

aexpect/rss_client.py
```diff
@@ -234,7 +246,7 @@ def _receive_file_chunks(self, filename, timeout=60):
         finally:
             f.close()
 
-    def _send_msg(self, msg, timeout=60):
+    def _send_msg(self, msg, _timeout=60):
```
ldoktor: Again, please delete the `timeout` or use `pylint: disable=`. Also in this case it'd probably make sense to implement the timeout :wink:

aexpect/remote.py
```diff
@@ -0,0 +1,1297 @@
...
+    :param path: List or string
+    :return: Shell escaped version
+    """
+    if isinstance(path, list):
+        return ' '.join(map(pipes.quote, path))
+    else:
+        return pipes.quote(path)
+
+
+def handle_prompts(session, username, password, prompt, timeout=10,
```
ldoktor: It'd be nice to (later) add default for prompt, what do you think?

ldoktor: ... and another idea, how about introducing PROMPT_LINUX and PROMPT_WINDOWS global variables as hints?

aexpect/remote.py
```diff
@@ -0,0 +1,1297 @@
+"""
+Functions and classes used for logging into guests and transferring files.
+"""
```
ldoktor: One more thing, the `remote` and `rss_client` should be listed in `aexpect.__init__`.

ldoktor:
Oh, one more thing (I know there was more). Please write a simple unittest. It doesn't have to be fancy, you can simply fake the RemoteSession and keep returning the expected strings. One positive + one negative should do for now as the feature itself should be tested in virttest already (but feel free to add complete testing as well...)

aexpect/remote.py
```diff
@@ -75,6 +75,20 @@ def __str__(self):
         return "%s    (output: %r)" % (self.msg, self.output)
 
 
+class TransferBadClientError(TransferError):
```
ldoktor: Looking at this hierarchy, the `msg` and `output` doesn't really scale here. How about introducing simple `RemoteError(Exception):  pass` and inherit from that? This Exception would be then used also in https://github.com/avocado-framework/aexpect/pull/55/commits/3a25533c4f7a4f21d0f9464e8f8ee20b6b8d993c#r291033819 as no output is available there...

aexpect/remote.py
```diff
@@ -75,6 +75,20 @@ def __str__(self):
...
+
+    def __init__(self, client):
+        TransferError.__init__(self, None, None)
+        self.client = client
+
+    def __str__(self):
+        return "Unknown file copy client: '%s', valid values are scp and rss" % self.client
+
+
+class SCPError(TransferError):
```
ldoktor: Do we need the per-protocol exception granularity?

aexpect/remote.py
```diff
@@ -11,15 +11,8 @@
...
 import aexpect
-from avocado.core import exceptions
-from avocado.utils import process
-
-from virttest import data_dir
 from virttest import utils_misc
 from virttest import rss_client
-from virttest.remote_commander import remote_master
-from virttest.remote_commander import messenger
-from virttest.compat_52lts import results_stdout_52lts, results_stderr_52lts
```
ldoktor: I don't use these and they are fairly advanced. I think they can stay in virttest only so yes to this commit.

aexpect/remote.py
```diff
@@ -469,7 +471,7 @@ def remote_scp(command, password_list, log_filename=None, transfer_timeout=600,
     logging.debug("Trying to SCP with command '%s', timeout %ss",
                   command, transfer_timeout)
     if log_filename:
-        output_func = utils_misc.log_line
+        output_func = log_function
```
ldoktor: Well this changes the behaviour as one might specify the `log_filename` as well as `log_function` and expect it to be logged on both places. Similarly setting the `log_filename` and without `log_function` would raise an exception, while before it did exactly what is in docstring.

If you really need to specify the log_function I'd suggest turning the `log_filename` into 3-state:

* str - write to file/log/stream
* callable - call it
* None - do nothing

alternatively we could keep both of them, but then the docstring has to mention they are mutually exclusive and when `log_filename` is specified it should use the internal implementation.

ldoktor: PS: I haven't tested it so I could be wrong...

ldoktor: One more thing I noticed, you need to add the arguments as last, otherwise it'd break existing scripts that use positional arguments. Another reason for 3-state.

Let me summarize. I'd prefer:

1. not to introduce log_function
2. introduce it as tri-state via `remote_scp(command, password_list, log_filename=None, transfer_timeout=600, login_timeout=300, log=None):` where `log_filename` would be deprecated, log warning on use and `log` would accept the 3 types
3. introduce the `log_function` but keep backward compatibility: `remote_scp(command, password_list, log_filename=None, transfer_timeout=600, login_timeout=300, log_function=None):`, using `log_filename` and `log_function` should raise exception and `log_filename` should use internal implementatio nof `log_line`.

aexpect/client.py
```diff
@@ -1352,6 +1352,77 @@ def get_command_status(self, cmd, timeout=60, internal_timeout=None,
...
+                              output_func, output_params, output_prefix, thread_name,
+                              prompt, status_test_command,
+                              pass_fds, encoding)
+
+        # Remember some attributes
+        self.client = client
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
```
ldoktor: It'd be nice to see `copy_from` and `copy_to` methods here...

ldoktor:
Last but not least, all travis failures need to be tackled. Let me know whether you need any help there. It's acceptable to locally disable some via `pylint: disable=` comment (for example the too-many-arguments, perhaps it could be done even globally in this file), but most of them should be resolved by refactoring and code improvements as they are there for a reason.

aexpect/remote.py
pevogam: Yes, makes perfect sense. I was doing too many other things when naming these functions. Resolved in the next push.

aexpect/remote.py
pevogam: I am not sure what you mean by the license "has to be here" since it is already added in the LGTM fixes (last) commit. Thanks for the sample link, I have added

```diff
...
# See LICENSE for more details.
+#
+# This code was imported from the avocado-vt project,
+#
+# virttest/remote.py

"""
Functions and classes used for logging into guests and transferring files.
...
```

However, there was no mentioning of the original author in either avocado-vt/virttest/remote.py or the virt-test. How do we find out who is the original author? The above addition is all I know about the original code.

aexpect/remote.py
pevogam: Good point, added in the next push. However, I will not resolve this until you confirm that the change is indeed exactly what you meant.

aexpect/remote.py
pevogam: Good, so I assume no action required on my side, just keeping this note open and not resolving it. Correct me if I am wrong but I leave it as is.

aexpect/rss_client.py
pevogam: Could it be that you have some cache which didn't update correctly? This license is available as well. I just double checked at the "Files" from the menu in this PR.

pevogam: We have author here, so I added this, we only don't have sources for authors for the `remote.py` utility.

aexpect/remote.py
pevogam: Sure, resolved with the next push.

aexpect/remote.py
pevogam: I would say it's better to have it for all of them for better safety and symmetry. Please, unresolve this note if you don't agree.

aexpect/remote.py
pevogam: I still wonder what is the use of this argument anyway but I will just rely on `pyling: disable=unused-argument` for now.

aexpect/remote.py
lmr: As far as I recall the original author for remote.py is Michael Goldish <mgoldish@redhat.com> at the time. I guess some git repo digging could be done, if my historical recollection is not enough :)

aexpect/rss_client.py
pevogam: Hehe sure, I am just sticking to the minimum modification policy right now. Even modifications of the smallest size usually require separate commits and are not part of the pure migration process + risk breaking more things for more people and require much more testing. I have made the argument usable for the next push.

aexpect/remote.py
pevogam: I agree, also the second comment is a really good idea. Both are implemented in the next push.

aexpect/remote.py
pevogam: Indeed, in order for us to best utilize resources I would say that if someone is using them very intensely and they want them in aexpect, they can always raise an issue to request this or contribute with their own migration as they might already know this code better as well.

pevogam:
Hi @ldoktor,

> There are some little things that needs to be changed, important are the licenses that should be there since the beginning including the mentioning of the original author and source. The rest is negotiable. Just don't forget you promised to eventually follow-up in virttest (after this is merged).

The licenses are already there, I just added authorship where some was available. The virttest-side PR is already ready for review and I referred to it in https://github.com/avocado-framework/aexpect/pull/55#issuecomment-496456310.

> Oh, one more thing (I know there was more). Please write a simple unittest. It doesn't have to be fancy, you can simply fake the RemoteSession and keep returning the expected strings. One positive + one negative should do for now as the feature itself should be tested in virttest already (but feel free to add complete testing as well...)

I did not find any relevant unit tests that I could migrate from the virt-test side - the only unit tests written were about some old file-related classes that I left there due to stronger dependencies. Maybe we can add this in a separate PR? I commit to it of course, it is just that this branch is becoming >6 commits large and I would really prefer to keep the commit granularity and not squash irrelevant or migration-violating commits.

> Last but not least, all travis failures need to be tackled. Let me know whether you need any help there. It's acceptable to locally disable some via `pylint: disable=` comment (for example the too-many-arguments, perhaps it could be done even globally in this file), but most of them should be resolved by refactoring and code improvements as they are there for a reason.

Yes, as I already mentioned in the second paragraph of https://github.com/avocado-framework/aexpect/pull/55#issuecomment-498957945 the only leftover problems are related to missing docstrings (agreed this is bad), short variable names (not necessarily wrong as in 'except Exception as e' idioms), and overcomplicated functions with too many branches, too many lines, or too many arguments. To be honest, especially the last types of problems cannot be resolved in any way other than large scale refactoring of the code itself which also becomes too much modification for a migration-dedicated PR like this one. I would suggest that we disable these overcomplicating complaints (I agree with them to some extend) for now until we decide to refactor this functionality in the future. The only complaints that I find acceptable right now are some due to missing docstrings (in which case I can only hope I understand each function right enough to document it).

aexpect/remote.py
pevogam: I think we already have a good granularity of protocols when it comes to different login/transfer operations so I see no reason not to have this granularity also for exception handling. If people don't bother about it, they can always catch the more general exception.

aexpect/remote.py
pevogam: Good point, corrected with the next push. Please take another look before I resolve this issue so that we make sure the exception hierarchy matches your expectations fully.

aexpect/client.py
pevogam: I was thinking the same but the problem is that this is yet another extension of the class since it will require specifying `file_transfer_client`, `file_transfer_port`, and a few file-transfer-specific parameters in addition to the basic ones for a remote session. Also, as I mentioned before copying files will require creation of a new file-transfer-type-of session and having `copy_from/to` as nonstatic method will only make sense if the methods are similar to the ones I added as a sample to `remote.py` like `scp_from_session` and `scp_to_session`.

aexpect/remote.py
pevogam: I see, not providing the `log_function` explicitly would imply that we have to provide it myself as before and migrate some code from `utils_misc` as well. I am against tristates or other magical multitypes which can lead to a lot of future headache. The third option is also plausible but first I will look into a way to port the additional code like `utils_misc.get_log_filename` and `utils_misc.log_line`.

pevogam: Hmmm, one major reason I left the log function implementation to the caller is that `utils_misc` is doing some strange magic with its implementation of `log_line` making use of globals and lots of things that will both complicate the configuration of aexpect and require heavier migration on the side of all avocado-vt. Providing just `log_file` xor `log_line` is also not possible since the aexpect side makes use of both in the implementation of `remote_scp` (thus raising exception if both are supplied no longer makes sense either). This leaves us only with part of your option 3 as a possibility. Any ideas for minimal and easily replaceable (e.g. by avocado-vt) internal `log_line` implementation so that we could default to option 1?

aexpect/remote.py
pevogam: I traced the module through avocado-vt and then virt-test and I think it starts somewhere in the autotest repo but I prefer to just trust the actual committer at these times ;)

pevogam: Author added to both modules now, resolving this.

aexpect/remote.py
ldoktor: sure, just asking :-)

pevogam:
Hi @ldoktor, just to clarify I am not working on any feature in this PR any longer. My comments above are waiting on your feedback but I am not planning to undertake any further steps just in case you are waiting on me.

pevogam:
Hi @ldoktor, it's been a few months and large scale modifying PR like this degrades very fast with time which is why it would be great if we can act on it faster. I have already addressed all of your issues above in https://github.com/avocado-framework/aexpect/pull/55#issuecomment-499455919 and the review comments. I am aware that you are very limited on resources which is why I wanted to offer more of my own help on the maintenance side if you would like me to. I could then gradually add more pull requests addressing the issues postponed from this PR and take full responsibility (in terms of quick reaction and hotfixes) in case of any incompatible changes shall the need arise.

ldoktor:
Hello @pevogam, I'm sorry I totally forgot about this PR. In case of full backlog I am only looking at PRs that pass CI, which is not the case of this one, so next time it helps to tide-up the PR.

Anyway your comment https://github.com/avocado-framework/aexpect/pull/55#issuecomment-499455919 is right. It'd be nice to have more selftests, but we can start with a manual check.

So the only missing issues are the docstrings and other pylint issues. I suggested you can use `pylint: disable=...` so please do so and add the docstrings everywhere.

Yes I'm prepared to turn a blind eye and have this function in an unstable version, but it must not break the CI.

ldoktor:
Let me summarize once again, what is required for this to be included:

1. CI must pass
2. docstrings must make sense
3. there should be example of sample usage (at least something like https://github.com/avocado-framework/aexpect/pull/55#pullrequestreview-246390904 ) either as a selftest or at least as a docstring/commented section in the module.
4. the example must work as well as other functional tests I'll run (but it already works, just mentioning it here :D)
5. unless you are sure the API is stable there must be a note about the limited support at the beginning of the module (something like EXPERIMENTAL or so, big enough to scare people away unless they know what are they doing).

provided all those are fulfilled I'll merge it so you can follow-up with Avocado-vt and potential improvements here.

pevogam:
Hi @ldoktor

> Hello @pevogam, I'm sorry I totally forgot about this PR. In case of full backlog I am only looking at PRs that pass CI, which is not the case of this one, so next time it helps to tide-up the PR.

This is not possible for a PR where we discuss actual CI issues and what way we should go about them. I also pinged you directly a few times just in case.

> Anyway your comment [#55 (comment)](https://github.com/avocado-framework/aexpect/pull/55#issuecomment-499455919) is right. It'd be nice to have more selftests, but we can start with a manual check.
> 
> So the only missing issues are the docstrings and other pylint issues. I suggested you can use `pylint: disable=...` so please do so and add the docstrings everywhere.

Alright, done in the most recent version after some extra push iterations.

> Yes I'm prepared to turn a blind eye and have this function in an unstable version, but it must not break the CI.

CI is green now :)

> Note a `:warning: experimental module` or so at the beginning of the `aexpect.remote` module would make sense to me....

AFAIR the API hasn't changed from the migration and every move we have made here is purely extensive, i.e. adding new functionality without modifying old functionaly.

pevogam:
> Let me summarize once again, what is required for this to be included:
> 
>     1. CI must pass
> 
>     2. docstrings must make sense
> 
>     3. there should be example of sample usage (at least something like [#55 (review)](https://github.com/avocado-framework/aexpect/pull/55#pullrequestreview-246390904) ) either as a selftest or at least as a docstring/commented section in the module.
> 
>     4. the example must work as well as other functional tests I'll run (but it already works, just mentioning it here :D)
> 
>     5. unless you are sure the API is stable there must be a note about the limited support at the beginning of the module (something like EXPERIMENTAL or so, big enough to scare people away unless they know what are they doing).
> 
> 
> provided all those are fulfilled I'll merge it so you can follow-up with Avocado-vt and potential improvements here.

To the best of my expectation each part of the summery should now be satisfied. There is only a slight requirement to update the python3 dependencies as was done by the separate PR #61 which should be merged before we can merge this one.

pevogam:
Pushed again just to rebase on top of python3 fixed master.

aexpect/remote.py
```diff
@@ -334,8 +334,10 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
 
     if verbose:
         logging.debug("Login command: '%s'", cmd)
-    session = ShellSession(cmd, linesep=linesep, prompt=prompt,
-                           status_test_command=status_test_command)
+    session = RemoteSession(cmd, linesep=linesep, prompt=prompt,
+                            status_test_command=status_test_command,
+                            client=client, host=host, port=port,
+                            username=username, password=password)
```
ldoktor: Yep, this should be harmless

aexpect/remote.py
```diff
@@ -1,14 +1,35 @@
...
+# Original author: Michael Goldish <mgoldish@redhat.com>
+#
+# Copyright: 2016 IBM
+# Authors : Michael Goldish <mgoldish@redhat.com>
+
 """
 Functions and classes used for logging into guests and transferring files.
 """
+
+# disable too-many-* as we need them pylint: disable=R0912,R0913,R0914,R0915,C0302
```
ldoktor: sure, no problem

pevogam: It imitates other modules for the time being ^^

aexpect/remote.py
pevogam: +1

aexpect/remote.py
ldoktor: Looking at it again (because I forgot I already have) it is not a regexp, it is just incorrect string. It should be `"-P %s %s %s@\\[%s\\]:%s"` as we want to put there `\[` and not to escape `[`. The result is the same, but it'd be better to use pure strings to avoid confusion in the future, anyway it's not a high priority right now...

aexpect/remote.py
```diff
@@ -17,6 +17,12 @@
 from aexpect import rss_client
 
 
+#: prompt to be used for shell sessions on linux machines (default)
+PROMPT_LINUX = r"^\\[.*\\][\\#\\$]\\s*$"
+#: prompt to be used for shell sessions on windows machines
+PROMPT_WINDOWS = r"^\w:\\.*>\s*$"
```
ldoktor: Useful

aexpect/remote.py
```diff
@@ -17,6 +17,12 @@
 from aexpect import rss_client
 
 
+#: prompt to be used for shell sessions on linux machines (default)
+PROMPT_LINUX = r"^\\[.*\\][\\#\\$]\\s*$"
```
ldoktor: Here you probably combined multiple escapes and regexp. If you want to write it as a regexp, you should use single `\`, if you want to write it as string, use multiple `\\`, but this does not work for me. Tested by:

```
>>> a = aexpect.remote.remote_login("ssh", "192.168.122.162", 22, "root", "redhat", aexpect.remote.PROMPT_LINUX, verbose=True)
Traceback (most recent call last):
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/remote.py", line 266, in handle_prompts
    timeout=timeout, internal_timeout=0.5)
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/client.py", line 947, in read_until_last_line_matches
    print_func)
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/client.py", line 866, in read_until_output_matches
    raise ExpectTimeoutError(patterns, output)
aexpect.exceptions.ExpectTimeoutError: Timeout expired while looking for patterns ['[Aa]re you sure', '[Pp]assword:\\s*', '\\(or (press|type) Control-D to continue\\):\\s*$', '[Gg]ive.*[Ll]ogin:\\s*$', '(?<![Ll]ast )[Ll]ogin:\\s*$', '[Cc]onnection.*closed', '[Cc]onnection.*refused', '[Pp]lease wait', '[Ww]arning', '[Ee]nter.*username', '[Ee]nter.*password', '[Cc]onnection timed out', '^\\\\[.*\\\\][\\\\#\\\\$]\\\\s*$', 'Escape character is.*']    (output: '[root@localhost ~]# ')
>>> a = aexpect.remote.remote_login("ssh", "192.168.122.162", 22, "root", "redhat", "^\\[.*\\][\\#\\$]\\s*$", verbose=True)
>>> 
```

aexpect/remote.py
```diff
@@ -19,6 +19,27 @@
...
+
+::
+
+    import aexpect
+    import aexpect.remote
+
+    b = aexpect.ShellSession("sh")
+    b.cmd("echo asdf > /tmp/bbb")
+
+    a = aexpect.remote.remote_login("ssh", "127.0.0.1", 5000, "root", "123456", "[\\#\\$]")
```
ldoktor: It'd be nice to use the built-in `aexpect.remote.PROMPT_LINUX`

aexpect/remote.py
```diff
@@ -19,6 +19,27 @@
...
+    a.cmd("rm /tmp/aaa -Rf")
+
+    aexpect.remote.scp_to_reusing_session(a, "/tmp/bbb", "/tmp/aaa")
```
ldoktor: this method does not exists, only `scp_to_session`...

aexpect/remote.py
```diff
@@ -19,6 +19,27 @@
...
+    aexpect.remote.scp_from_reusing_session(a, "/tmp/aaa", "/tmp/ccc")
+
+    a.close()
+    b.close()
```
ldoktor: a usage with `with` would be probably better...

aexpect/remote.py
pevogam: Yes, it originates from earlier discussions one year ago and I think it was originally your suggestion (cannot recall any longer).

ldoktor:
Hello @pevogam it's almost there:

1. would you please remind me the resolution of https://github.com/avocado-framework/aexpect/pull/55/commits/ab46ec997bdd07e46c8b19d15f04c70ea552e5d5#r291050606 it's changing the API (moves some arguments further in their position) which would make the simultaneous usage of both `virttest` and `aexpect` versions harder as people frequently depend on the argument position.
2. not super important, but the https://github.com/avocado-framework/aexpect/pull/55/commits/56802bbad762698fd2e1d551e2abf0bb25bf6ab3#r374257126 would be a nice touch (using `-P %s %s %s@\\[%s\\]:%s` instead of the `r` prefixes)
3. Please fix the `PROMPT_LINUX` https://github.com/avocado-framework/aexpect/pull/55/commits/d55d50020bf3a32343bef7688031748eb0b062da#r374266441
4. Please update the example of usage according to notes from https://github.com/avocado-framework/aexpect/pull/55/commits/d55d50020bf3a32343bef7688031748eb0b062da

I have tested this functionally and it seems to work well (tested linux guests only)

aexpect/remote.py
pevogam: Good catch! Fixed.

aexpect/remote.py
pevogam: Done

aexpect/remote.py
pevogam: Hmmm I copy-pasted the example you provided previously wondering if it could be added to the docstring, I guess I took it too literally. Fixed now.

aexpect/remote.py
pevogam: Yes, it will also make the usage blocks of each session clearer - I only copy pasted a previous example without taking a deeper look into the example. Added now.

pevogam:
Hi @ldoktor,

>     1. would you please remind me the resolution of [ab46ec9#r291050606](https://github.com/avocado-framework/aexpect/commit/ab46ec997bdd07e46c8b19d15f04c70ea552e5d5#r291050606) it's changing the API (moves some arguments further in their position) which would make the simultaneous usage of both `virttest` and `aexpect` versions harder as people frequently depend on the argument position.

Yes, I agree and I am fully aware of the general requirement for API stability but in this case I am also handling the migration from the avocado-vt/virttest side and the code there has already been corrected for this. This means that people using virttest would not feel the difference unless they explicitly call functions like `remote_login` in which case they will immediately get a quick to fix error of mismatched function type. The argument position for the log function is most meaningful next to the log file and I anticipate very few direct calls to this function out there which is enough to leverage the bold step of putting the argument at the right place for a better long term readability. What do you think?

>     2. not super important, but the [56802bb#r374257126](https://github.com/avocado-framework/aexpect/commit/56802bbad762698fd2e1d551e2abf0bb25bf6ab3#r374257126) would be a nice touch (using `-P %s %s %s@\\[%s\\]:%s` instead of the `r` prefixes)

I think the `r` prefixes make more sense here since the resulting regex is easier to read (compared to extra escaped backslashes) and this is already the approach that was undertaken at the PROMPT constants in the beginning of the module (where it is also more readable this way).

>     3. Please fix the `PROMPT_LINUX` [d55d500#r374266441](https://github.com/avocado-framework/aexpect/commit/d55d50020bf3a32343bef7688031748eb0b062da#r374266441)

Yes done as per your comments above.

>     4. Please update the example of usage according to notes from [d55d500](https://github.com/avocado-framework/aexpect/commit/d55d50020bf3a32343bef7688031748eb0b062da)
> 

Also done as per your comments above.

ldoktor:
And I'm sorry it took so long. Remote commander was always tricky to deal with and not many people are willing to review it. But it does the job...

pevogam:
Yes, I think you are right that there might be a few additional tweaks we can add on both sides so let's see what follows. Thanks a lot for your time though, I know larger PRs like this are also much harder to review, yet ironically also easier to break if it takes more time. Will proceed on the other side now.

PR #56, opened 2019-05-04
Resolve any alerts generated by LGTM

pevogam:

### DIFF
--- a/scripts/aexpect_helper
+++ b/scripts/aexpect_helper
@@ -132,7 +132,7 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
         # Read from child and write to files/pipes
         server_log.info('Entering main read loop')
-        status = 1
+        status = 1  # lgtm [py/multiple-definition]
         while True:
             check_termination = False
             # Make a list of reader pipes whose buffers are not empty


ldoktor:
Hello Plamen, do you have a link with the LGTM failure? Theoretically it shouldn't be necessary to set the `status`. I'm just afraid someone might add another `break` without setting the `status` that would introduce a failure (that should get caught by CI, but you never know...). I'd prefer to keep the default value there just to be safe unless it breaks our checks. I tried pylint and it is perfectly happy scoring 10/10.

ldoktor:
OK I see the failure: https://lgtm.com/projects/g/avocado-framework/aexpect/snapshot/a8fb3109f6971461585c98297800d7fca3603b6b/files/scripts/aexpect_helper?sort=name&dir=ASC&mode=heatmap#x19d9afa53ba4e242:1

I tried pylint and it won't catch this issue, nor it catches extra `break` without status setting, which means only LGTM would potentially recognize such failure.

pevogam:
Hi Lukáš,

> OK I see the failure: https://lgtm.com/projects/g/avocado-framework/aexpect/snapshot/a8fb3109f6971461585c98297800d7fca3603b6b/files/scripts/aexpect_helper?sort=name&dir=ASC&mode=heatmap#x19d9afa53ba4e242:1
> 
> I tried pylint and it won't catch this issue, nor it catches extra `break` without status setting, which means only LGTM would potentially recognize such failure in the future.

So what is your preferred course of action? Should we fix this to make LGTM (and future pull requests passing through LGTM checks) happy or leave it as is for extra safety? I would opt in for the first option but your call is more relevant here.

ldoktor:
What about this:

```diff
diff --git a/scripts/aexpect_helper b/scripts/aexpect_helper
index c9614b4..da7d881 100755
--- a/scripts/aexpect_helper
+++ b/scripts/aexpect_helper
@@ -132,7 +132,6 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
         # Read from child and write to files/pipes
         server_log.info('Entering main read loop')
-        status = 1
         while True:
             check_termination = False
             # Make a list of reader pipes whose buffers are not empty
@@ -170,10 +169,7 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
                     buffers[i] += data
             # If os.read() raised an exception or there was nothing to read --
             if check_termination or shell_fd not in r_fds:
-                pid, status = os.waitpid(shell_pid, os.WNOHANG)
-                if pid:
-                    status = os.WEXITSTATUS(status)
-                    break
+                break
             # If there's data to read from the client --
             if inpipe_fd in r_fds:
                 data = os.read(inpipe_fd, 1024)
@@ -181,6 +177,9 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
         server_log.info('Out of the main read loop. Writing status to %s',
                         status_filename)
+        pid, status = os.waitpid(shell_pid, os.WNOHANG)
+        if pid:
+            status = os.WEXITSTATUS(status)
         with open(status_filename, "w") as file_obj:
             file_obj.write(str(status))
```

ldoktor:
Oups, that won't work...

ldoktor:
@pevogam how about simply adding the `lgtm [py/multiple-definition]` at the end of the line to skip this check for this particular line? I'd prefer that as without lgtm we are blind to possible future changes like:

```diff
diff --git a/scripts/aexpect_helper b/scripts/aexpect_helper
index c9614b4..1f5b82c 100755
--- a/scripts/aexpect_helper
+++ b/scripts/aexpect_helper
@@ -132,7 +132,6 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
         # Read from child and write to files/pipes
         server_log.info('Entering main read loop')
-        status = 1
         while True:
             check_termination = False
             # Make a list of reader pipes whose buffers are not empty
@@ -169,6 +168,8 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
                 for i in range(len(readers)):
                     buffers[i] += data
             # If os.read() raised an exception or there was nothing to read --
+            if check_termination:
+                break
             if check_termination or shell_fd not in r_fds:
                 pid, status = os.waitpid(shell_pid, os.WNOHANG)
                 if pid:
```

pevogam:
> @pevogam how about simply adding the `lgtm [py/multiple-definition]` at the end of the line to skip this check for this particular line? I'd prefer that as without lgtm we are blind to possible future changes like:
> 

Good point, just force pushed with the ignore line comment right now.

PR #58, opened 2019-06-03
Fix fallout from exception catch refactoring in fdff994efd

pevogam:
The safer cmd_output call disregards ExpectTimeoutError exceptions
and sends an extra new line to clear the shell prompt in case of
unwanted ouput which made it fail previously matching the same prompt.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1204,8 +1204,7 @@ class ShellSession(Expect):
                 success = True
                 break
             except ExpectTimeoutError as error:
-                output = self.remove_command_echo(error.output, cmd)
-                raise ShellTimeoutError(cmd, output)
+                self.sendline()
             except ExpectProcessTerminatedError as error:
                 output = self.remove_command_echo(error.output, cmd)
                 raise ShellProcessTerminatedError(cmd, error.status, output)


ldoktor:
As a matter of fact a while ago I replaced the safe commands for normal ones as it was misbehaving. I should have investigated it more thoroughly...

pevogam:
> As a matter of fact a while ago I replaced the safe commands for normal ones as it was misbehaving. I should have investigated it more thoroughly...

At least in our cases we have some tests making use of the safer command and it works fine. This tests could not pass with the normal command so the safe command was a good extension guess on the side of the original developer.

ldoktor:
Well, same story for me so when it started failing I replaced it with multiple attempts. Not proud of it, but unfortunately there was no time to look for details. I should give this another try...

pevogam:
> Well, same story for me so when it started failing I replaced it with multiple attempts. Not proud of it, but unfortunately there was no time to look for details. I should give this another try...

Yes definitely. I had to fix it in order to make our own tests pass and multiple attempts would not have helped since we are dealing with a command that adds kernel output directly into the session. I hope it is fixed for you too then.

PR #63, opened 2020-09-22
Preparations for release 1.6.0

clebergnu:
Cleanups and CI fixes in preparation for release 1.6.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: python
 
 python:
-    - "2.7"
     - "3.5"
     - "3.6"
     - "3.7"
+    - "3.8"
 
 sudo: false
 
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ rpm-release: srpm-release
 	mock -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "rel_build 1" --rebuild BUILD/SRPM/python-aexpect-$(VERSION)-*.src.rpm
 
 check: clean
-	inspekt checkall --disable-lint W1618,R0205
+	inspekt checkall --disable-lint W1618,R0205,W0707,R1725
 	$(PYTHON) setup.py develop $(PYTHON_DEVELOP_ARGS)
 	$(PYTHON) setup.py test
 
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -576,10 +576,12 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
     # main retrieval of the local object
     import importlib
     module = importlib.import_module(object_name)
+
     def proxymethod(fun):
         """Decorator to autoproxy all callables."""
         def wrapper(*args, **kwargs):
             rarg = fun(*args, **kwargs)
+
             def proxify_type(rarg):
                 if rarg is None or type(rarg) in (bool, int, float, str):  # pylint: disable=C0123
                     return rarg
@@ -599,6 +601,7 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
                 return generator_wrapper()
             return proxify_type(rarg)
         return wrapper
+
     class ModuleObject:  # pylint: disable=R0903
         """Module wrapped for transferability."""
     for fname, fobj in inspect.getmembers(module, inspect.isfunction):
@@ -749,6 +752,7 @@ def import_remote_exceptions(exceptions):
         exceptiontype = RemoteCustomException
         exception = exceptiontype(*class_dict["args"])
         # in the case of non-custom exceptions the class is properly restored
+        # pylint: disable=W0201
         exception.__customclass__ = class_dict.get("__class__", "")
 
         if "attributes" in class_dict.keys():
--- a/aexpect/utils/path.py
+++ b/aexpect/utils/path.py
@@ -72,7 +72,7 @@ def init_dir(*args):
     :return: directory.
     :rtype: str
     """
-    directory = os.path.join(*args)
+    directory = os.path.join(*args)  # pylint: disable=E1120
     if not os.path.isdir(directory):
         os.makedirs(directory)
     return directory
--- a/setup.py
+++ b/setup.py
@@ -15,17 +15,10 @@
 
 """Aexpect setup script"""
 
-import sys
-# pylint: disable=E0611
-
 from setuptools import setup
 
 
 if __name__ == '__main__':
-    if sys.version_info[0] == 2:
-        REQUIREMENTS = ['subprocess32>=3.2.6']
-    else:
-        REQUIREMENTS = []
     setup(name='aexpect',
           version='1.5.1',
           description='Aexpect',
@@ -35,6 +28,4 @@ if __name__ == '__main__':
           packages=['aexpect',
                     'aexpect.utils'],
           scripts=['scripts/aexpect_helper'],
-          use_2to3=True,
-          install_requires=REQUIREMENTS,
           test_suite='tests')
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -182,16 +182,7 @@ class Spawn(object):
 
         # Start the server (which runs the command)
         if command:
-            # try to find python specific version of aexpect_helper first, then
-            # try unversioned
-            helper_noversion = 'aexpect_helper'
-            helper_versioned = '{0}-{1}.{2}'.format(helper_noversion,
-                                                    sys.version_info[0],
-                                                    sys.version_info[1])
-            try:
-                helper_cmd = utils_path.find_command(helper_versioned)
-            except utils_path.CmdNotFoundError:
-                helper_cmd = utils_path.find_command(helper_noversion)
+            helper_cmd = utils_path.find_command('aexpect_helper')
             sub = subprocess.Popen([helper_cmd],
                                    shell=True,
                                    stdin=subprocess.PIPE,
--- a/python-aexpect.spec
+++ b/python-aexpect.spec
@@ -10,10 +10,10 @@
 %global gittar		%{srcname}-%{version}.tar.gz
 %else
 %if ! 0%{?commit:1}
-%global commit		012bc0e57fc9d05d2ee4ca2425dc27b07b745a32
+%global commit		be9be4b325ac1da7b0c908e82f1d2c52e43dfd2f
 %endif
 %if ! 0%{?commit_date:1}
-%global commit_date	20180613
+%global commit_date	20200922
 %endif
 %global shortcommit	%(c=%{commit};echo ${c:0:7})
 %global gitrel		.%{commit_date}git%{shortcommit}
@@ -23,23 +23,9 @@
 # Selftests are provided but skipped because they use unsupported tooling.
 %global with_tests 0
 
-%if 0%{?rhel}
-    %if 0%{?rhel} == 7
-        %global with_python2 1
-        %global with_python3 0
-    %else
-        # RHEL 8 and later
-        %global with_python2 0
-        %global with_python3 1
-    %endif
-%else
-    %global with_python2 1
-    %global with_python3 1
-%endif
-
 Name: python-%{srcname}
 Version: 1.5.1
-Release: 1%{?gitrel}%{?dist}
+Release: 2%{?gitrel}%{?dist}
 Summary: Aexpect is a python library to control interactive applications
 Group: Development/Tools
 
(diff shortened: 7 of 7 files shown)


pevogam:
image
:tada:

.travis.yml
```diff
@@ -1,10 +1,10 @@
 language: python
 
 python:
-    - "2.7"
```
pevogam, line 4: I am not sure if cutting 2.7 will block the removal of the remote module on the Avocado VT side. There was a proposal by @sathnaga and a few other maintainers in the past to drop support of 2.7 there and move on but I am not sure about the state of this proposal.

.travis.yml
```diff
@@ -1,10 +1,10 @@
...
     - "3.5"
     - "3.6"
     - "3.7"
+    - "3.8"
```
pevogam, line 7: Awesome!

pevogam (approved):
If it depends on me this is a LGTM, let's see what the others think about the Python 2.7 situation.

Makefile
```diff
@@ -73,7 +73,7 @@ rpm-release: srpm-release
 	mock -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "rel_build 1" --rebuild BUILD/SRPM/python-aexpect-$(VERSION)-*.src.rpm
 
 check: clean
-	inspekt checkall --disable-lint W1618,R0205
+	inspekt checkall --disable-lint W1618,R0205,W0707,R1725
```
ldoktor, line 76: Is there a reason to add these? I tried running it with W0707 and R1725 enabled and it passed. I'd really prefer addressing the actual places only.

aexpect/remote_door.py
```diff
@@ -749,6 +752,7 @@ def recreate_exception(class_name, class_dict):
         exceptiontype = RemoteCustomException
         exception = exceptiontype(*class_dict["args"])
         # in the case of non-custom exceptions the class is properly restored
+        # pylint: disable=W0201
```
ldoktor, line 755: How about simply adding the `__customclass__` attribute on line 749 with the value None?

aexpect/utils/path.py
```diff
@@ -72,7 +72,7 @@ def init_dir(*args):
     :return: directory.
     :rtype: str
     """
-    directory = os.path.join(*args)
+    directory = os.path.join(*args)  # pylint: disable=E1120
```
ldoktor, line 75: which version is failing? It works well on my system without the disable.

.travis.yml
ldoktor: Unless it's breaking anything I'd also prefer to keep aexpect at least buildable on python 2.7. Is there a strict reason to abandon it?

clebergnu: > Unless it's breaking anything I'd also prefer to keep aexpect at least buildable on python 2.7. Is there a strict reason to abandon it?

There's current breakage wrt Python 2.7 syntax compatibility, for instance:

```
Compiling /builddir/build/BUILDROOT/python-aexpect-1.5.1-1.20200525gitbe9be4b.fc31.x86_64/usr/lib/python2.7/site-packages/aexpect/remote_door.py ...
  File "/usr/lib/python2.7/site-packages/aexpect/remote_door.py", line 127
    def run_remote_util(session, utility, function, *args, verify=None, detach=False, **kwargs):
                                                                ^
SyntaxError: invalid syntax
```

Makefile
clebergnu: See: https://travis-ci.org/github/clebergnu/aexpect/jobs/729403637#L263 and https://travis-ci.org/github/clebergnu/aexpect/jobs/729403637#L274

aexpect/remote_door.py
clebergnu: I went with the least intrusion approach... I did not want to introduce behavior change in a "prepare to release / get CI green" kind of PR.

aexpect/utils/path.py
clebergnu: > which version is failing? It works well on my system without the disable.

This is what I'm getting:

```
************* Module aexpect.utils.path
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
```

This is with:

```
$ pylint --version
pylint 2.5.0
astroid 2.4.2
Python 3.7.9 (default, Aug 19 2020, 17:05:11) 
[GCC 9.3.1 20200408 (Red Hat 9.3.1-2)]

$ inspekt --version
inspekt 0.5.2
```

.travis.yml
pevogam: Yes, we had this breakage at the time of the PR and decided not to go around it but let the CI fail. I guess you have a point that this concerns building as well though. I guess worst case we can drop the remote door module through a bit of magic for a 2.7 SPEC even though the changes dropping it from the SPEC were already done and the progress there will be lost. I am still not sure why keeping Python 2.7 is still so relevant. @ldoktor Do you mind explaining some of the use cases for Python 2.7 here? I guess I have to ask the same in relation to Avocado VT where you previously insisted we should also keep such compatibility. If is is just a few users relying on this could they maybe clone the repos in their previous version to keep using them?

aexpect/remote_door.py
pevogam: Yes, and as we already have a few pylint comments like this left to resolve in the future, we can just consider this as one of them as well. It is best to test before doing greater changes than this.

pevogam: I could add more unit tests in the future but right now this should be enough for the release.

setup.py
```diff
@@ -15,17 +15,10 @@
...
 
-import sys
-# pylint: disable=E0611
-
 from setuptools import setup
 
 
 if __name__ == '__main__':
-    if sys.version_info[0] == 2:
-        REQUIREMENTS = ['subprocess32>=3.2.6']
```
luckyh, line 26: Looks we need to modify the following part accordingly?

https://github.com/avocado-framework/aexpect/blob/be9be4b325ac1da7b0c908e82f1d2c52e43dfd2f/aexpect/client.py#L55-L58

.travis.yml
ldoktor: The reason for me is users. Yes, they can checkout oldish versions, but it's easier to just keep it working unless it becames a big burden. Addressing the issue should be quite simple as we are receiving full kwargs.

Anyway (similarly to Avocado-vt) if the fix becomes too ugly or extensive, I won't block py2 removal.

As for the py2 usage, it's been in use for old-rhel on new-rhel testing. You are right, it can be replaced with oldish Avocado-vt in guests, but these do not contain useful fixes and maintaining multiple branches downstream (for each team as teams usually only cooperate upstream) is not as convenient as keeping py2 alive.

Makefile
ldoktor: I see. If we decide to drop py2 we should also address all these (especially the W0707) as they require py3. If we kept py2 skipping these is fine.

aexpect/remote_door.py
ldoktor: My problem is that I spent a lot of time fixing and polishing aexpect. Saying "left to resolve in the future" without a name and a deadline doesn't seem valid to me. Adding the attribute on class-level should be more explicit.

aexpect/utils/path.py
ldoktor: Could you please try updating your pylint? It looks like a bug to me:

```
$ inspekt checkall --disable-lint W1618,R0205,W0707,R1725
PEP8 disabled: E501,E265,W601,E402
Pylint disabled: W1618,R0205,W0707,R1725
Pylint enabled : W0611
License check: (gplv2_later)
Global check PASS
[medic@localhost aexpect ]$ pylint --version
pylint 2.6.0
astroid 2.4.2
Python 3.8.5 (default, Aug 12 2020, 00:00:00) 
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)]
[medic@localhost aexpect ]$ inspekt --version
inspekt 0.5.2
```

aexpect/remote_door.py
pevogam: @ldoktor I don't see why "left to resolve in the future" is a problem. Everybody uses the same words here, this is the same way the aexpect release was specified to me - as something that will happen "soon". But in contrast to the fact that it has already taken more than half a year, you have already seen that I execute what I say and in a timely manner. In fact, my greatest problem with aexpect so far has been not about me taking too much time when I commit to a task but to me committing to a task that then blocked me from progressing for another year.

Considering the breakage last minutes changes could create and the fact the code in its current version is already in use, I would prefer the first option.

.travis.yml
pevogam: I think all of this makes sense. The question that remains to answer together then is where to draw the line of too complex and expensive. I don't mean here specifically but in the long run. I guess this could pass the bar.

aexpect/remote_door.py
pevogam: Btw, I already have a PR with some small improvements and some extra unit tests coming with these that I haven't pushed considering that aexpect is already overwhelmed with a few PRs with new features and we were waiting for a release in order to be able to finish the migration on the Avocado VT side first before taking further steps here.

.travis.yml
clebergnu: > The reason for me is users. Yes, they can checkout oldish versions, but it's easier to just keep it working unless it becames a big burden. Addressing the issue should be quite simple as we are receiving full kwargs.
> 
> Anyway (similarly to Avocado-vt) if the fix becomes too ugly or extensive, I won't block py2 removal.
> 
> As for the py2 usage, it's been in use for old-rhel on new-rhel testing. You are right, it can be replaced with oldish Avocado-vt in guests, but these do not contain useful fixes and maintaining multiple branches downstream (for each team as teams usually only cooperate upstream) is not as convenient as keeping py2 alive.

Python 2 is D-E-A-D. Users on Python 2 can get older versions. I've talked to heavy users of Avocado + Avocado-VT + aexpect on the Python 2 situation (EL 7), and they have their versions pinned for compatibility.

@ldoktor if you find this important enough to send fixes and maintain Python 2 support at all times, then we can keep it.  If not, it must go.

Makefile
clebergnu: > I see. If we decide to drop py2 we should also address all these (especially the W0707) as they require py3. If we kept py2 skipping these is fine.

Are you volunteering? :)

aexpect/utils/path.py
clebergnu: > Could you please try updating your pylint? It looks like a bug to me:
> 
> ```
> $ inspekt checkall --disable-lint W1618,R0205,W0707,R1725
> PEP8 disabled: E501,E265,W601,E402
> Pylint disabled: W1618,R0205,W0707,R1725
> Pylint enabled : W0611
> License check: (gplv2_later)
> Global check PASS
> [medic@localhost aexpect ]$ pylint --version
> pylint 2.6.0
> astroid 2.4.2
> Python 3.8.5 (default, Aug 12 2020, 00:00:00) 
> [GCC 10.2.1 20200723 (Red Hat 10.2.1-1)]
> [medic@localhost aexpect ]$ inspekt --version
> inspekt 0.5.2
> ```

I'm not sure what's going on with either mine or your system, but this is on a fresh venv with pylint 2.6.0:

```
$ make check 
/tmp/pylint260/bin/python setup.py clean
running clean
make -f /home/cleber/src/avocado/aexpect/debian/rules clean || true
make[1]: Entering directory '/home/cleber/src/avocado/aexpect'
/home/cleber/src/avocado/aexpect/debian/rules:7: /usr/share/cdbs/1/class/python-distutils.mk: No such file or directory
make[1]: *** No rule to make target '/usr/share/cdbs/1/class/python-distutils.mk'.  Stop.
make[1]: Leaving directory '/home/cleber/src/avocado/aexpect'
rm -rf build/ MANIFEST BUILD BUILDROOT SPECS RPMS SRPMS SOURCES
find . -name '*.pyc' -delete
inspekt checkall --disable-lint W1618,R0205
PEP8 disabled: E501,E265,W601,E402
Pylint disabled: W1618,R0205
Pylint enabled : W0611
License check: (gplv2_later)
/home/cleber/src/avocado/aexpect/aexpect/remote_door.py:579:5: E306 expected 1 blank line before a nested definition, found 0
/home/cleber/src/avocado/aexpect/aexpect/remote_door.py:583:13: E306 expected 1 blank line before a nested definition, found 0
/home/cleber/src/avocado/aexpect/aexpect/remote_door.py:602:5: E306 expected 1 blank line before a nested definition, found 0
Python library autopep8 not installed. Please install it if you want to use --fix
Style: /home/cleber/src/avocado/aexpect/aexpect/remote_door.py FAIL (FIX NOT OK)
************* Module aexpect.utils.path
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
E1120: 75,16: init_dir: No value for argument 'a' in function call
************* Module aexpect.remote_door
W0201:752,8: import_remote_exceptions.recreate_exception: Attribute '__customclass__' defined outside __init__
make: *** [Makefile:76: check] Error 6
(pylint260) [cleber@localhost aexpect]$ which pylint
/tmp/pylint260/bin/pylint
(pylint260) [cleber@localhost aexpect]$ pylint --version
pylint 2.6.0
astroid 2.4.2
Python 3.7.9 (default, Aug 19 2020, 17:05:11) 
[GCC 9.3.1 20200408 (Red Hat 9.3.1-2)]
```

aexpect/remote_door.py
clebergnu: > My problem is that I spent a lot of time fixing and polishing aexpect. Saying "left to resolve in the future" without a name and a deadline doesn't seem valid to me. Adding the attribute on class-level should be more explicit.

I understand the effort you put in, but please consider that we're addressing the current state: CI is broken, and we're possibly not catching *other issues*. Also, we're attempting to do a release.

I've created an issue to track the overall pylint status: https://github.com/avocado-framework/aexpect/issues/64

clebergnu:
Closing in favor of https://github.com/avocado-framework/aexpect/pull/65

PR #66, opened 2020-09-25
Makefile: add a pypi target to ease future releases

clebergnu:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,5 +83,18 @@ clean:
 	rm -rf build/ MANIFEST BUILD BUILDROOT SPECS RPMS SRPMS SOURCES
 	find . -name '*.pyc' -delete
 
+pypi: clean
+	if test ! -d PYPI_UPLOAD; then mkdir PYPI_UPLOAD; fi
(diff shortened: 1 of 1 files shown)


pevogam (commented):
How about fully automating the upload? This is how we do it for the Avocado I2N plugin:

https://github.com/intra2net/avocado-i2n/blob/e12349ebb0b5810edc1020aeeb748a6f0d0e2bf2/.travis.yml#L62

ldoktor (commented):
Would you please add the PYPI_UPLOAD dir to the cleanup? Also depends whether you want to use @pevogam's automation suggestion...

lmr:
That automation looks nice. I'll take a look at it as well.

lmr:
Merged, thanks @clebergnu !

pevogam:
@lmr This PR needed to add the `PYPI_UPLOAD` dir to the cleanup as @ldoktor suggested or to consider an even easier approach first. I guess we were in a hurry to automate it or something, not sure.

PR #67, opened 2020-09-26
Remote door improvements

pevogam:
Extension of the remote door to remove Windows machines,
added support for indented functions, output returning remote
function calls, auto-imported deserializable exceptions.

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -712,9 +712,11 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
     :param int port: port of the remote sharing server
     :param str os_type: OS type of the session, either "linux" or "windows"
     :param extra_params: extra parameters to pass to the remote object sharing
-                         control file (similarly to subcontrol setting above),
+                         control file (sismilarly to subcontrol setting above),
                          with keys usually prepended with "ro_" prefix
     :type extra_params: {str, str}
+    :returns: newly created middleware session for the remote object server
+    :rtype: :py:class:`RemoteSession`
     :raises: :py:class:`RuntimeError` if the object server failed to start
 
     In comparison to :py:func:`share_local_object`, this function fires up a
@@ -729,6 +731,7 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
     .. note:: Created and works specifically for Windows and Linux.
     """
     logging.info("Sharing the remote objects over the network")
+    extra_params = {} if extra_params is None else extra_params
 
     # setup remote objects server
     logging.info("Starting nameserver for the remote objects")
@@ -743,11 +746,12 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
     # optional parameters (set only if present and/or available)
     for key in extra_params.keys():
         local_path = set_subcontrol_parameter(local_path, key, extra_params[key])
+    remote_path = os.path.join(REMOTE_CONTROL_DIR,
+                               os.path.basename(control_path))
     # NOTE: since we are creating the path in Linux but use it in Windows,
     # we replace some of the backslashes
     if os_type == "windows":
-        remote_path = os.path.join(REMOTE_PYTHON_PATH,
-                                   os.path.basename(control_path)).replace("/", "\\")
+        remote_path = remote_path.replace("/", "\\")
     remote.copy_files_to(session.host, transfer_client,
                          session.username, session.password, transfer_port,
                          local_path, remote_path, timeout=10)
@@ -772,6 +776,7 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
 
     Pyro4.config.NS_HOST = host
     logging.getLogger("Pyro4").setLevel(10)
+    return middleware_session
 
 
 def import_remote_exceptions(exceptions=None, modules=None):
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -185,6 +185,10 @@ def run_remotely(function):
         logging.debug("Running remotey a function using the wrapper control %s",
                       wrapper_control)
 
+        # fix line endings for Windows clients
+        if session.client == "nc":
+            wrapper_control = wrapper_control.replace("\n", "\r\n")
+
         control_path = _string_generated_control(wrapper_control)
         run_subcontrol(session, control_path)
 
@@ -194,6 +198,27 @@ def run_remotely(function):
 ##################################
 #   STATIC (TEMPLATE) CONTROLS   #
 ##################################
+
+
+def _copy_control(session, control_path, is_utility=False):
+    remote_dir = REMOTE_PYTHON_PATH if is_utility else REMOTE_CONTROL_DIR
+    # run on remote Linux hosts
+    if session.client == "ssh":
+        transfer_client = "scp"
+        transfer_port = 22
+        remote_control_path = os.path.join(remote_dir, os.path.basename(control_path))
+    # run on remote Windows hosts
+    elif session.client == "nc":
+        transfer_client = "rss"
+        transfer_port = 10023
+        # ..todo:: use `remote_dir` here
+        remote_control_path = "%TEMP%\\" + os.path.basename(control_path)
+    else:
+        raise NotImplementedError("run_subcontrol not implemented for client %s" % session.client)
+    remote.copy_files_to(session.host, transfer_client,
+                         session.username, session.password, transfer_port,
+                         control_path, remote_control_path)
+    return remote_control_path
 
 
 def run_subcontrol(session, control_path, timeout=600, detach=False):
@@ -207,10 +232,18 @@ def run_subcontrol(session, control_path, timeout=600, detach=False):
     :param bool detach: whether to detach from session (e.g. if running a
                         daemon or similar long-term utility)
     """
-    remote_control_path = os.path.join(REMOTE_CONTROL_DIR, os.path.basename(control_path))
-    remote.scp_to_remote(session.host, session.port, session.username, session.password,
-                         control_path, remote_control_path)
-    cmd = REMOTE_PYTHON_BINARY + " " + remote_control_path
+    remote_control_path = _copy_control(session, control_path)
+    # run on remote Linux hosts
+    if session.client == "ssh":
+        python_binary = REMOTE_PYTHON_BINARY
+    # run on remote Windows hosts
+    # ..todo:: combine with REMOTE_PYTHON_BINARY
+    elif session.client == "nc":
+        python_binary = session.cmd("where python", timeout=timeout,
+                                    print_func=logging.info).strip()
+    else:
+        raise NotImplementedError("run_subcontrol not implemented for client %s" % session.client)
+    cmd = python_binary + " " + remote_control_path
     if detach:
         session.set_output_func(logging.info)
         session.set_output_params(())
@@ -460,8 +493,7 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
             raise
 
         # if there is no door on the other side, open one
-        remote.scp_to_remote(session.host, session.port, session.username, session.password,
-                             os.path.abspath(__file__), REMOTE_PYTHON_PATH)
+        _copy_control(session, os.path.abspath(__file__), is_utility=True)
         run_remote_util(session, "remote_door", "share_local_object",
                         object_name, host=host, port=port, detach=True)
         output, attempts = "", 10
@@ -507,8 +539,7 @@ def get_remote_objects(session=None, host="localhost", port=0):
             raise
 
         # if there is no door on the other side, open one
-        remote.scp_to_remote(session.host, session.port, session.username, session.password,
-                             os.path.abspath(__file__), REMOTE_PYTHON_PATH)
+        _copy_control(session, os.path.abspath(__file__), is_utility=True)
         run_remote_util(session, "remote_door", "share_local_objects",
                         wait=True, host=host, port=port, detach=True)
         control_log = session.cmd("cat " + REMOTE_CONTROL_LOG)
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -61,6 +61,7 @@ import os
 import re
 import logging
 import inspect
+import importlib
 import threading
 import tempfile
 import time
@@ -610,7 +611,6 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
         nsd_running = False
 
     # main retrieval of the local object
-    import importlib
     module = importlib.import_module(object_name)
 
     def proxymethod(fun):
@@ -768,19 +768,40 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
     logging.getLogger("Pyro4").setLevel(10)
 
 
-def import_remote_exceptions(exceptions):
+def import_remote_exceptions(exceptions=None, modules=None):
     """
     Make serializable all remote custom exceptions.
 
-    :param exceptions: module dot accessible exception names
-    :type exceptions: [str]
+    :param exceptions: full module path exception names (optional)
+    :type exceptions: [str] or None
+    :param modules: full module paths whose custom exceptions will first be
+                    detected and then automatically imported (optional)
+    :type exceptions: [str] or None
+
+    The deserialization by our Pyro backend requires the full module paths to
+    each exception or module in order to correctly detect the exception type.
 
     .. note:: This wouldn't be needed if we were using the Pickle serializer but its
         security problems at the moment made us prefer the serpent serializer paying
         for it with some extra setup steps and this method.
     """
-    # ..todo:: need a more dynamic way of determining the exceptions like examining
-    # the errors modules and even extracting the exception classes from there
+    def list_module_exceptions(modstr):
+        module = importlib.import_module(modstr)
+        exceptions = []
+        for name in module.__dict__:
+            if not inspect.isclass(module.__dict__[name]):
+                continue
+            if (issubclass(module.__dict__[name], Exception) or name.endswith('Error')):
+                exceptions.append(modstr + "." + name)
+        return exceptions
+
+    exceptions = [] if not exceptions else exceptions
+    modules = [] if not modules else modules
+    for module in modules:
+        exceptions += list_module_exceptions(module)
+    logging.debug("Registering the following exceptions for deserialization: %s",
+                  ", ".join(exceptions))
+
     class RemoteCustomException(Exception):
         """Standard class to instantiate during remote expection deserialization."""
     def recreate_exception(class_name, class_dict):
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -115,7 +115,22 @@ def _string_call(function, *args, **kwargs):
     return "result = %s(%s)\n" % (function, arguments)
 
 
-def _string_generated_control(control_str):
+def _string_generated_control(client, control_body):
+    control_header = "import logging\n"
+    control_header += "logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '\n"  # pylint: disable=C0301
+    control_header += "                    'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')\n"  # pylint: disable=C0301
+    control_header += "import sys\n"
+    control_header += "sys.path.append('%s')\n" % REMOTE_PYTHON_PATH
+
+    # the string call will make sure this variable always exists
+    control_footer = "print('RESULT = ' + str(result))\n"
+
+    control_str = control_header + control_body + control_footer
+
+    # fix line endings for Windows clients
+    if client == "nc":
+        control_str = control_str.replace("\n", "\r\n")
+
     dump_dir = os.path.abspath(DUMP_CONTROL_DIR)
     new_fd, new_path = tempfile.mkstemp(suffix=".control", dir=dump_dir)
     logging.debug("Using %s for generated control file", new_path)
@@ -124,7 +139,7 @@ def _string_generated_control(control_str):
     return new_path
 
 
-def run_remote_util(session, utility, function, *args, verify=None, detach=False, **kwargs):
+def run_remote_util(session, utility, function, *args, detach=False, **kwargs):
     """
     Access a remote utility by generating a small control file to run remotely.
 
@@ -132,11 +147,10 @@ def run_remote_util(session, utility, function, *args, verify=None, detach=False
     :type session: RemoteSession object
     :param str utility: name of the remote utility to run
     :param str function: function of the remote utility to call
-    :param verify: expected return argument from the call to use for
-                   verification of the call execution
-    :type verify: object or None
     :param bool detach: whether to detach from session (e.g. if running a
                         daemon or similar long-term utility)
+    :returns: serialized return argument from the call
+    :rtype: str
 
     If the remote utility call is detached from the session, it will not wait
     for the session command to complete. Its further output (stdout and stderr)
@@ -145,20 +159,17 @@ def run_remote_util(session, utility, function, *args, verify=None, detach=False
     communication from the remote utility in order to avoid polluting its
     output from various detached processes. This is not a problem as multiple
     sessions can be easily created.
-    """
-    wrapper_control = "import logging\n"
-    wrapper_control += "logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '\n"
-    wrapper_control += "                    'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')\n"
-    wrapper_control += "import sys\n"
-    wrapper_control += "sys.path.append('%s')\n" % REMOTE_PYTHON_PATH
-    wrapper_control += "import %s\n" % utility
-    wrapper_control += _string_call(utility + "." + function, *args, **kwargs)
-    wrapper_control += "assert result is %s\n" % verify if verify else ""
 
-    control_path = _string_generated_control(wrapper_control)
+    This supports return arguments of the decorated function but needs its own
+    deserialization since all arguments are returned as strings.
+    """
+    control_body = "import %s\n" % utility
+    control_body += _string_call(utility + "." + function, *args, **kwargs)
+    control_path = _string_generated_control(session.client, control_body)
     logging.debug("Accessing %s remote utility using the wrapper control %s",
                   utility, control_path)
-    run_subcontrol(session, control_path, detach=detach)
+    full_output = run_subcontrol(session, control_path, detach=detach)
+    return "None" if detach else re.search("RESULT = (.*)", full_output).group(1)
 
 
 def run_remotely(function):
@@ -172,32 +183,25 @@ def run_remotely(function):
 
     Each function should contain a `_session` as a first argument
     and is expected to follow PEP8 standards.
+
+    This supports return arguments of the decorated function but needs its own
+    deserialization since all arguments are returned as strings.
     """
     def wrapper(session, *args, **kwargs):
         # drop first argument and first line with the decorator since function already runs remotely
         fn_source = inspect.getsourcelines(function)[0][1:]
         indent_number = len(fn_source[0]) - len(fn_source[0].lstrip())
-
         # remove extra indentation from the beginning to allow indented functions
         if indent_number > 0:
             fn_source = [line[indent_number:] for line in fn_source]
 
-        wrapper_control = "import logging\n"
-        wrapper_control += "logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '\n"  # pylint: disable=C0301
-        wrapper_control += "                    'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')\n"  # pylint: disable=C0301
-        wrapper_control += "import sys\n"
-        wrapper_control += "sys.path.append('%s')\n" % REMOTE_PYTHON_PATH
-        wrapper_control += "\n" + "".join(fn_source).replace("_session, ", "")
-        wrapper_control += "\n" + _string_call(function.__name__, *args, **kwargs)
-        logging.debug("Running remotey a function using the wrapper control %s",
-                      wrapper_control)
-
-        # fix line endings for Windows clients
-        if session.client == "nc":
-            wrapper_control = wrapper_control.replace("\n", "\r\n")
-
-        control_path = _string_generated_control(wrapper_control)
-        run_subcontrol(session, control_path)
+        control_body = "\n" + "".join(fn_source).replace("_session, ", "")
+        control_body += "\n" + _string_call(function.__name__, *args, **kwargs)
+        control_path = _string_generated_control(session.client, control_body)
+        logging.debug("Running remotely a function using the wrapper control %s",
+                      control_path)
+        full_output = run_subcontrol(session, control_path)
+        return re.search("RESULT = (.*)", full_output).group(1)
 
     return wrapper
 
@@ -238,6 +242,8 @@ def run_subcontrol(session, control_path, timeout=600, detach=False):
     :param int timeout: timeout for the control to complete
     :param bool detach: whether to detach from session (e.g. if running a
                         daemon or similar long-term utility)
+    :returns: the full raw output from the call or empty string if detached
+    :rtype: str
     """
     remote_control_path = _copy_control(session, control_path)
     # run on remote Linux hosts
@@ -255,8 +261,8 @@ def run_subcontrol(session, control_path, timeout=600, detach=False):
         session.set_output_func(logging.info)
         session.set_output_params(())
         session.sendline(cmd)
-    else:
-        session.cmd(cmd, timeout=timeout, print_func=logging.info)
+        return ""
+    return session.cmd(cmd, timeout=timeout, print_func=logging.info)
 
 
 def prep_subcontrol(src_file, src_dir=None):
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -75,7 +75,13 @@ except ImportError:
     logging.warning("Remote object backend (Pyro4) not found, some functionality"
                     " of the remote door will not be available")
 
-from aexpect import remote
+# NOTE: disable aexpect importing on the remote side if not available as the
+# remote door can run code remotely without the requirement for the aexpect
+# module, alas, offering just limited functionality
+try:
+    from aexpect import remote
+except ImportError:
+    pass
 
 
 #############################################
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -480,12 +480,12 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
     If `session` is not `None`, you will not be able to use it after this call
     since it is reserved for communication with the remote object. For example,
     do not run::
-        vm, session = vmnet.get_single_vm_with_session()
+        session = remote.wait_for_login(...)
         my_util = get_remote_object('package.module', session, ...)
         session.cmd('ls')   # will time out
 
     but instead::
-        vm, session = vmnet.get_single_vm_with_session()
+        session = remote.wait_for_login(...)
         my_util = get_remote_object('package.module', vm.wait_for_login(), ...)
         session.cmd('ls')   # works now
 
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -96,22 +96,14 @@ REMOTE_PYTHON_PATH = "/tmp/utils"
 
 
 def _string_call(function, *args, **kwargs):
-    arguments = ""
-    for arg in args:
+    def arg_to_str(arg):
         if isinstance(arg, str):
-            arguments = "%sr'%s'" % (arguments, arg)
-        else:
-            arguments = "%s%s" % (arguments, arg)
-        if len(kwargs) > 0 or arg != args[-1]:
-            arguments = "%s%s" % (arguments, ", ")
-    ordered_kwargs = list(kwargs.keys())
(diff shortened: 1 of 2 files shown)


aexpect/remote_door.py
```diff
@@ -174,13 +174,20 @@ def run_remotely(function):
     and is expected to follow PEP8 standards.
     """
     def wrapper(session, *args, **kwargs):
+        # drop first argument and first line with the decorator since function already runs remotely
+        fn_source = inspect.getsourcelines(function)[0][1:]
+        indent_number = len(fn_source[0]) - len(fn_source[0].lstrip())
+
+        # remove extra indentation from the beginning to allow indented functions
+        if indent_number > 0:
+            fn_source = [line[indent_number:] for line in fn_source if len(line) > indent_number]
```
ldoktor, line 186: Do you have a valid reason for the `if len(line) > indent_number`? If it's not longer than not including it or including `line[indent_number:]` should result in the same (``). (not a strict thing just wondering)

aexpect/remote_door.py
```diff
@@ -174,13 +174,20 @@ def run_remotely(function):
...
+
         wrapper_control = "import logging\n"
         wrapper_control += "logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '\n"  # pylint: disable=C0301
         wrapper_control += "                    'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')\n"  # pylint: disable=C0301
         wrapper_control += "import sys\n"
         wrapper_control += "sys.path.append('%s')\n" % REMOTE_PYTHON_PATH
-        # drop first argument and first line with the decorator since function already runs remotely
-        wrapper_control += "\n" + "\n".join(inspect.getsource(function).split("\n")[1:]).replace("_session, ", "")  # pylint: disable=C0301
+        wrapper_control += "\n" + "".join(fn_source).replace("_session, ", "")
```
ldoktor, line 193: it'd be probably nice to expand the selftest. How about this:

```diff
diff --git a/tests/test_remote_door.py b/tests/test_remote_door.py
index 07facee..297b2fb 100644
--- a/tests/test_remote_door.py
+++ b/tests/test_remote_door.py
@@ -23,12 +23,6 @@ import unittest.mock as mock
 from aexpect import remote_door
 
 
-@remote_door.run_remotely
-def add_one(number):
-    """A small decorated test function."""
-    return number + 1
-
-
 @mock.patch('aexpect.remote_door.remote', mock.MagicMock())
 class RemoteDoorTest(unittest.TestCase):
     """Unit test class for the remote door."""
@@ -84,6 +78,14 @@ class RemoteDoorTest(unittest.TestCase):
 
     def test_run_remote_decorator(self):
         """Test that a remote utility object can be run properly."""
+        @remote_door.run_remotely
+        def add_one(number):
+            """A small decorated test function."""
+            def foo():
+
+                pass
+
+            return number + 1
         _ = add_one(self.session, 3)  # pylint: disable=E1121
         # python 3.6 and above
         if hasattr(self.session.cmd, "assert_called_once"):
@@ -99,6 +101,8 @@ class RemoteDoorTest(unittest.TestCase):
             self.assertIn("def add_one(number):\n", control_lines)
             self.assertIn("    return number + 1\n", control_lines)
             self.assertIn("result = add_one(3)\n", control_lines)
+            self.assertIn("    def foo():\n", control_lines)
+            self.assertIn("        pass\n", control_lines)
 
     def test_run_remote_object(self):
         """Test that a remote utility object can be run properly."""
```

aexpect/remote_door.py
```diff
@@ -242,8 +242,8 @@ def run_subcontrol(session, control_path, timeout=600, detach=False):
         session.set_output_func(logging.info)
         session.set_output_params(())
         session.sendline(cmd)
-    else:
-        session.cmd(cmd, timeout=timeout, print_func=logging.info)
+        return None
+    return session.cmd(cmd, timeout=timeout, print_func=logging.info)
```
ldoktor, line 249: Makes sense, what would you say about using mock to replace `remote.copy_files_to` from line `242` in `remote_door` for `shutil.copy` (or open&write the local file) and using an actual `aexpect.ShellSession("sh")` for the selftest? (which is how I actually tested this on my machine using the `test_remote_door.py`)

aexpect/remote_door.py
```diff
@@ -104,7 +104,7 @@ def _string_call(function, *args, **kwargs):
             arguments = "%s%s" % (arguments, arg)
         if len(kwargs) > 0 or arg != args[-1]:
             arguments = "%s%s" % (arguments, ", ")
-    ordered_kwargs = list(kwargs.keys())
+    ordered_kwargs = sorted(list(kwargs.keys()))
```
ldoktor, line 110: Perhaps it'd be nice to add a note here that the sorting is important for selftest purposes. Otherwise someone might optimize this out...

ldoktor: Actually the whole function seems unnecessary complex, how about this:

```diff
diff --git a/aexpect/remote_door.py b/aexpect/remote_door.py
index c1c603e..2b5e017 100644
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -99,22 +99,15 @@ REMOTE_PYTHON_PATH = "/tmp/utils"
 
 
 def _string_call(function, *args, **kwargs):
-    arguments = ""
-    for arg in args:
+    def arg_to_str(arg):
         if isinstance(arg, str):
-            arguments = "%sr'%s'" % (arguments, arg)
+            return "r'%s'" % arg
         else:
-            arguments = "%s%s" % (arguments, arg)
-        if len(kwargs) > 0 or arg != args[-1]:
-            arguments = "%s%s" % (arguments, ", ")
-    ordered_kwargs = sorted(list(kwargs.keys()))
-    for key in ordered_kwargs:
-        if isinstance(kwargs[key], str):
-            arguments = "%s%s=r'%s'" % (arguments, key, kwargs[key])
-        else:
-            arguments = "%s%s=%s" % (arguments, key, kwargs[key])
-        if key != ordered_kwargs[-1]:
-            arguments = "%s%s" % (arguments, ", ")
+            return "%s" % arg
+    args = tuple(arg_to_str(arg) for arg in args)
+    kwargs = tuple("%s=%s" % (key, arg_to_str(value))
+                   for key, value in sorted(kwargs.items()))
+    arguments = ", ".join(args + kwargs)
     return "result = %s(%s)\n" % (function, arguments)
 
```

Or do you think I should send this separately as it's not related to this change?

aexpect/remote_door.py
```diff
@@ -759,19 +759,29 @@ def share_remote_objects(session, control_path, host="localhost", port=9090,
...
-    # ..todo:: need a more dynamic way of determining the exceptions like examining
-    # the errors modules and even extracting the exception classes from there
+    def list_module_exceptions(mod):
+        module = __import__(mod)
+        exceptions = []
+        for name in module.__dict__:
+            if (issubclass(module.__dict__[name], Exception) or name.endswith('Error')):
+                exceptions.append(name)
+        return exceptions
+    exceptions = list_module_exceptions(module) if module else exceptions
```
ldoktor, line 786: I can't say I completely understand the `import_remote_exceptions` method, but this addition which just detects all exceptions makes sense. Would be nice to add a selftest for this. IIUC it should be possible even on a single host, right?

aexpect/remote_door.py
```diff
@@ -74,7 +74,10 @@
     logging.warning("Remote object backend (Pyro4) not found, some functionality"
                     " of the remote door will not be available")
 
-from aexpect import remote
+try:
+    from aexpect import remote
+except ImportError:
+    pass
```
ldoktor, line 80: Yes, it should work, but again it'd be nice to add a comment here to explain why are we permitting this.

aexpect/remote_door.py
```diff
@@ -465,12 +465,12 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
...
     since it is reserved for communication with the remote object. For example,
     do not run::
-        vm, session = vmnet.get_single_vm_with_session()
+        session = remote.wait_for_login(...)
         my_util = get_remote_object('package.module', session, ...)
         session.cmd('ls')   # will time out
 
     but instead::
-        vm, session = vmnet.get_single_vm_with_session()
+        session = remote.wait_for_login(...)
```
ldoktor, line 473: sure, vm should not be necessary for that :-)

aexpect/remote_door.py
```diff
@@ -207,10 +211,26 @@ def run_subcontrol(session, control_path, timeout=600, detach=False):
...
+        transfer_client = "rss"
+        transfer_port = 10023
+        # ..todo:: use REMOTE_CONTROL_DIR?
+        remote_control_path = "%TEMP%\\" + os.path.basename(control_path)
+        python_binary = session.cmd("where python", timeout=timeout,
+                                    print_func=logging.info).strip()
+    else:
+        raise NotImplementedError("run_subcontrol not implemented for client %s" % session.client)
+    remote.copy_files_to(session.host, transfer_client,
+                         session.username, session.password, transfer_port,
```
ldoktor, line 241: Yes, this should result in the same no linux and something that looks windowish on windows. Is the session.client always set for the usual sessions? (I'm only faking it when testing) If not than it'd perhaps be nice to detect it rather than to expect it exists.

ldoktor (commented):
It looks quite good and things that I tested (usually mocked or on localhost) worked well. Would you please add the comments so we don't forget about things and optimize things out in the future and also please consider updating/adding the selftests.

pevogam:
> It looks quite good and things that I tested (usually mocked or on localhost) worked well. Would you please add the comments so we don't forget about things and optimize things out in the future and also please consider updating/adding the selftests.

Thanks a lot for the review @ldoktor! I will take a look and accommodate everything in the beginning of the coming week.

aexpect/remote_door.py
pevogam: The session object is of type `RemoteSession` which [always sets the client attribute](https://github.com/avocado-framework/aexpect/blob/76c6019a3d8afa3b2cc4de3bcc269e076e1724ac/aexpect/client.py#L1405). I am still concerned a bit about the value hard-coding here and the unfinished `REMOTE_CONTROL_DIR` implementation but this PR has been waiting for quite long in my shelves and it still adds windows capabilities so I guess we can improve on this front gradually depending on demand for windows hosts.

aexpect/remote_door.py
pevogam: :blush:

aexpect/remote_door.py
pevogam: Good point, I usually describe such things in commit messages but always prefer to put them in the code base where they are more visible and last longer. I just need to be reminded from time to time :smile:, thanks!

aexpect/remote_door.py
pevogam: Absolutely! This little function started simple but then it grew in complexity through the years as we kept adding complications so I guess it reached its point of much needed refactoring. I applied your suggest changes with some credit in the commit message :)

aexpect/remote_door.py
pevogam: Good combination of localizing dependency functions like `add_one` and including indentation into the testing. All done.

aexpect/remote_door.py
pevogam: I undertook some steps so feel free to tell me if it reflects your points in the following commit (to be pushed):
```
commit dfec762c36dba01355bcc23ce114ab4b57a2703b (HEAD -> door-improvements)
Author: Plamen Dimitrov <pdimitrov@pevogam.com>
Date:   Mon Oct 5 13:10:00 2020 +0300

    Improved the coverage of remote door's unit tests
    
    In particular we:
    
    1) Include testing of indentation and nested functions.
    2) Make the remote mock replacements more explicit.
    3) Make use of a localized RemoteSession instance.
    
```

However, I am not sure what benefit we get from mocking `remote.copy_files_to` through `shutil.copy`. We are not validating against any copied files and only check the locally stored file for the right content.

aexpect/remote_door.py
pevogam: Thanks for noticing, it is ideal to batch a unit test together with the newer more complex functionality here. Done in the push I just did.

aexpect/remote_door.py
pevogam: I extended the windows functionality a bit leaving clearer todos regarding some further integration.

aexpect/remote_door.py
pevogam: Good point, I know that it is normal in `numpy` but wasn't sure here. It works fine as you suggested.

pevogam:
@ldoktor I pushed with some incompatibilities with Python 3.8- versions, will take a look during the second round of review (second reviewer or more feedback) since I am running out of time and I could batch it with more changes.

pevogam:
It is hard to predict when this PR will get another review so I used the time while everything is still fresh in my head to:

1) push a fix for all python versions of the exception serialization unit test
2) extend and do some additional functional testing of the same functionality

The PR is now final on my side unless there are more worthy change requests.

ldoktor:
Hello @pevogam, thank you for the updates, the fake localhost session was meant to switch from mocked selftests to actual selftests without the ssh. Something like:

```diff
diff --git a/tests/test_remote_door.py b/tests/test_remote_door.py
index 85477ce..a0c9d70 100644
--- a/tests/test_remote_door.py
+++ b/tests/test_remote_door.py
@@ -17,9 +17,11 @@
 import os
 import glob
 import re
+import shutil
 import unittest
 import unittest.mock as mock
 
+import aexpect
 from aexpect import remote
 from aexpect import remote_door
 
@@ -28,13 +30,20 @@ def _local_login(client, host, port, username, password, prompt, linesep="\n",
                  log_filename=None, log_function=None, timeout=10,
                  interface=None, identity_file=None,
                  status_test_command="echo $?", verbose=False, bind_ip=None):
-    return remote.RemoteSession("sh", linesep=linesep, prompt=prompt,
-                                status_test_command=status_test_command,
-                                client=client, host=host, port=port,
-                                username=username, password=password)
+    session = aexpect.ShellSession("sh")
+    session.client = "dummy"
+    session.host = "host"
+    session.port = 1234
+    session.username = "user"
+    session.password = "pass"
+    session.client = "ssh"
+    return session
 
+def _local_copy(host, client, username, password, port, src, dst):
+    import pydevd; pydevd.settrace("127.0.0.1", True, True)
+
+    shutil.copy(src, dst)
 
-@mock.patch('aexpect.remote_door.remote.copy_files_to', mock.MagicMock())
 @mock.patch('aexpect.remote_door.remote.remote_login', _local_login)
 class RemoteDoorTest(unittest.TestCase):
     """Unit test class for the remote door."""
@@ -88,6 +97,7 @@ class RemoteDoorTest(unittest.TestCase):
                           "r'Wonderland is fun', wanderer=None)\n",
                           control_lines)
 
+    @mock.patch('aexpect.remote_door.remote.copy_files_to', _local_copy)
     def test_run_remote_decorator(self):
         """Test that a remote decorated function can be run properly."""
         @remote_door.run_remotely
@@ -98,25 +108,15 @@ class RemoteDoorTest(unittest.TestCase):
                 pass
 
             do_nothing()
+            print("sample output")
             return number + 1
-        _ = add_one(self.session, 3)  # pylint: disable=E1121
-        # python 3.6 and above
-        if hasattr(self.session.cmd, "assert_called_once"):
-            self.session.cmd.assert_called_once()
-        else:
-            self.assertEqual(self.session.cmd.call_count, 1)
-        command = self.session.cmd.call_args[0][0]
-        self.assertTrue(re.match(r"python3 /tmp/tmp.+\.control", command),
-                        "A control file has to be generated and called on the peer")
-        control = os.path.basename(command.lstrip("python3 "))
-        with open(control) as handle:
-            control_lines = handle.readlines()
-            self.assertIn("def add_one(number):\n", control_lines)
-            self.assertIn("    def do_nothing():\n", control_lines)
-            self.assertIn("        pass\n", control_lines)
-            self.assertIn("    do_nothing()\n", control_lines)
-            self.assertIn("    return number + 1\n", control_lines)
-            self.assertIn("result = add_one(3)\n", control_lines)
+        import pydevd; pydevd.settrace("127.0.0.1", True, True)
+
+        session = _local_login(None, None, None, None, None, None)
+        ret = add_one(session, 3)  # pylint: disable=E1121
+        self.assertEqual(ret, "sample output\n")
+        # TODO: Check the control file output...
+        # TODO: Remove the contorl file
 
     def test_run_remote_object(self):
         """Test that a remote object can be run properly."""
```

As a bonus we'd get coverage for the `return` introduced in this series. I'd really want you to consider adding those functional selftests as it can decrease the amount of mocking in the selftests. If you decide not to do that please remove the additional mocks you added as they are not actually used and might give false impression that they are.

pevogam:
> Hello @pevogam, thank you for the updates, the fake localhost session was meant to switch from mocked selftests to actual selftests without the ssh. Something like:

Let me see since I have some questions regarding your suggestion here:

> ```diff
> diff --git a/tests/test_remote_door.py b/tests/test_remote_door.py
> index 85477ce..a0c9d70 100644
> --- a/tests/test_remote_door.py
> +++ b/tests/test_remote_door.py
> @@ -17,9 +17,11 @@
>  import os
>  import glob
>  import re
> +import shutil
>  import unittest
>  import unittest.mock as mock
>  
> +import aexpect
>  from aexpect import remote
>  from aexpect import remote_door
>  
> @@ -28,13 +30,20 @@ def _local_login(client, host, port, username, password, prompt, linesep="\n",
>                   log_filename=None, log_function=None, timeout=10,
>                   interface=None, identity_file=None,
>                   status_test_command="echo $?", verbose=False, bind_ip=None):
> -    return remote.RemoteSession("sh", linesep=linesep, prompt=prompt,
> -                                status_test_command=status_test_command,
> -                                client=client, host=host, port=port,
> -                                username=username, password=password)
> +    session = aexpect.ShellSession("sh")
> +    session.client = "dummy"
> +    session.host = "host"
> +    session.port = 1234
> +    session.username = "user"
> +    session.password = "pass"
> +    session.client = "ssh"
> +    return session
>```

This code sets `client` multiple times and I am not at clear why you suggest `ShellSession` given that
the remore door utility uses `RemoteSession` in particular.

> ```diff
> +def _local_copy(host, client, username, password, port, src, dst):
> +    import pydevd; pydevd.settrace("127.0.0.1", True, True)
> +
> +    shutil.copy(src, dst)
>```

Do we need the extra `pydevd` dependency for the tests? The `mock` import is standard dependency and
doesn't require external packages like this.

> ```diff
> -@mock.patch('aexpect.remote_door.remote.copy_files_to', mock.MagicMock())
>  @mock.patch('aexpect.remote_door.remote.remote_login', _local_login)
>  class RemoteDoorTest(unittest.TestCase):
>      """Unit test class for the remote door."""
>```

What is the reason for removing hte `copy_files_to` from the class? It is used by most unit tests and
not just by the `test_run_remote_decorator`.

> ```diff
> @@ -88,6 +97,7 @@ class RemoteDoorTest(unittest.TestCase):
>                            "r'Wonderland is fun', wanderer=None)\n",
>                            control_lines)
>  
> +    @mock.patch('aexpect.remote_door.remote.copy_files_to', _local_copy)
>      def test_run_remote_decorator(self):
>          """Test that a remote decorated function can be run properly."""
>          @remote_door.run_remotely
> @@ -98,25 +108,15 @@ class RemoteDoorTest(unittest.TestCase):
>                  pass
>  
>              do_nothing()
> +            print("sample output")
>```

What is the reason for the print statement? We are not checking logs or anything.

> ```diff
>              return number + 1
> -        _ = add_one(self.session, 3)  # pylint: disable=E1121
> -        # python 3.6 and above
> -        if hasattr(self.session.cmd, "assert_called_once"):
> -            self.session.cmd.assert_called_once()
> -        else:
> -            self.assertEqual(self.session.cmd.call_count, 1)
> -        command = self.session.cmd.call_args[0][0]
> -        self.assertTrue(re.match(r"python3 /tmp/tmp.+\.control", command),
> -                        "A control file has to be generated and called on the peer")
> -        control = os.path.basename(command.lstrip("python3 "))
> -        with open(control) as handle:
> -            control_lines = handle.readlines()
> -            self.assertIn("def add_one(number):\n", control_lines)
> -            self.assertIn("    def do_nothing():\n", control_lines)
> -            self.assertIn("        pass\n", control_lines)
> -            self.assertIn("    do_nothing()\n", control_lines)
> -            self.assertIn("    return number + 1\n", control_lines)
> -            self.assertIn("result = add_one(3)\n", control_lines)
> +        import pydevd; pydevd.settrace("127.0.0.1", True, True)
> +
> +        session = _local_login(None, None, None, None, None, None)
>```

What is the reason for using the login method like this? Each unit test is provided with a default
session it can use in `self.session` attribute.

> ```diff
> +        ret = add_one(session, 3)  # pylint: disable=E1121
> +        self.assertEqual(ret, "sample output\n")
>```

The `ret`-urn value should be "4" and not a print statement here.

> ```diff
> +        # TODO: Check the control file output...
>```

Why did you remove the code checking the control file output above?

> ```diff
> +        # TODO: Remove the contorl file
>```

Not sure what this TODO means.

> ```diff
>      def test_run_remote_object(self):
>          """Test that a remote object can be run properly."""
> ```
> 
> As a bonus we'd get coverage for the `return` introduced in this series.

Perhaps we could find a way to test the return value by preserving the mocks, I will take a look.

> I'd really want you to consider adding those functional selftests as it can decrease the amount of mocking in the selftests.

What is bad about mocking? I understand these tests as unit tests and less as functional tests, the focus
remaining on the tested utility and taken too far away into the overall integration of such.

> If you decide not to do that please remove the additional mocks you added as they are not actually used and might give false impression that they are.

Can you be a bit more specific which additional mocks do you mean? All the mocks I have added that I know of are actually used
so I am confused here.

ldoktor:
> > Hello @pevogam, thank you for the updates, the fake localhost session was meant to switch from mocked selftests to actual selftests without the ssh. Something like:
> 
> Let me see since I have some questions regarding your suggestion here:
> 
> > ```diff
> > diff --git a/tests/test_remote_door.py b/tests/test_remote_door.py
> > index 85477ce..a0c9d70 100644
> > --- a/tests/test_remote_door.py
> > +++ b/tests/test_remote_door.py
> > @@ -17,9 +17,11 @@
> >  import os
> >  import glob
> >  import re
> > +import shutil
> >  import unittest
> >  import unittest.mock as mock
> >  
> > +import aexpect
> >  from aexpect import remote
> >  from aexpect import remote_door
> >  
> > @@ -28,13 +30,20 @@ def _local_login(client, host, port, username, password, prompt, linesep="\n",
> >                   log_filename=None, log_function=None, timeout=10,
> >                   interface=None, identity_file=None,
> >                   status_test_command="echo $?", verbose=False, bind_ip=None):
> > -    return remote.RemoteSession("sh", linesep=linesep, prompt=prompt,
> > -                                status_test_command=status_test_command,
> > -                                client=client, host=host, port=port,
> > -                                username=username, password=password)
> > +    session = aexpect.ShellSession("sh")
> > +    session.client = "dummy"
> > +    session.host = "host"
> > +    session.port = 1234
> > +    session.username = "user"
> > +    session.password = "pass"
> > +    session.client = "ssh"
> > +    return session
> > ```
> 
> This code sets `client` multiple times and I am not at clear why you suggest `ShellSession` given that
> the remore door utility uses `RemoteSession` in particular.
> 

It's not a complete example, just something I put together during the review and yes, it needs to be polished. Yes, it should be RemoteSession, but when I was trying it out, it didn't work. The issue was not related to RemoteSession in the end, but after changing to ShellSession (just to be sure) I did not change it back.

So yes, it should be RemoteSession.

> > ```diff
> > +def _local_copy(host, client, username, password, port, src, dst):
> > +    import pydevd; pydevd.settrace("127.0.0.1", True, True)
> > +
> > +    shutil.copy(src, dst)
> > ```
> 
> Do we need the extra `pydevd` dependency for the tests? The `mock` import is standard dependency and
> doesn't require external packages like this.
> 

Of course, not, it's just a left-over from debugging.

> > ```diff
> > -@mock.patch('aexpect.remote_door.remote.copy_files_to', mock.MagicMock())
> >  @mock.patch('aexpect.remote_door.remote.remote_login', _local_login)
> >  class RemoteDoorTest(unittest.TestCase):
> >      """Unit test class for the remote door."""
> > ```
> 
> What is the reason for removing the `copy_files_to` from the class? It is used by most unit tests and
> not just by the `test_run_remote_decorator`.
> 

Again, it was interfering with the other mock and the fastest way to create a PoC was to remove it. Of course it'd have to be present for the other tests (unless we turn all of the tests into functional ones).

> > ```diff
> > @@ -88,6 +97,7 @@ class RemoteDoorTest(unittest.TestCase):
> >                            "r'Wonderland is fun', wanderer=None)\n",
> >                            control_lines)
> >  
> > +    @mock.patch('aexpect.remote_door.remote.copy_files_to', _local_copy)
> >      def test_run_remote_decorator(self):
> >          """Test that a remote decorated function can be run properly."""
> >          @remote_door.run_remotely
> > @@ -98,25 +108,15 @@ class RemoteDoorTest(unittest.TestCase):
> >                  pass
> >  
> >              do_nothing()
> > +            print("sample output")
> > ```
> 
> What is the reason for the print statement? We are not checking logs or anything.
> 

As you can see in the selftest, the "sample output" is the only thing that got returned when calling the `add_one` function. I also expected the `4` to be returned, but that did not happened. Perhaps I'm doing something wrong but running this selftest resulted in this behaviour.

> > ```diff
> >              return number + 1
> > -        _ = add_one(self.session, 3)  # pylint: disable=E1121
> > -        # python 3.6 and above
> > -        if hasattr(self.session.cmd, "assert_called_once"):
> > -            self.session.cmd.assert_called_once()
> > -        else:
> > -            self.assertEqual(self.session.cmd.call_count, 1)
> > -        command = self.session.cmd.call_args[0][0]
> > -        self.assertTrue(re.match(r"python3 /tmp/tmp.+\.control", command),
> > -                        "A control file has to be generated and called on the peer")
> > -        control = os.path.basename(command.lstrip("python3 "))
> > -        with open(control) as handle:
> > -            control_lines = handle.readlines()
> > -            self.assertIn("def add_one(number):\n", control_lines)
> > -            self.assertIn("    def do_nothing():\n", control_lines)
> > -            self.assertIn("        pass\n", control_lines)
> > -            self.assertIn("    do_nothing()\n", control_lines)
> > -            self.assertIn("    return number + 1\n", control_lines)
> > -            self.assertIn("result = add_one(3)\n", control_lines)
> > +        import pydevd; pydevd.settrace("127.0.0.1", True, True)
> > +
> > +        session = _local_login(None, None, None, None, None, None)
> > ```
> 
> What is the reason for using the login method like this? Each unit test is provided with a default
> session it can use in `self.session` attribute.

Again, it's remnant from trying this with ShellSession directly.

> 
> > ```diff
> > +        ret = add_one(session, 3)  # pylint: disable=E1121
> > +        self.assertEqual(ret, "sample output\n")
> > ```
> 
> The `ret`-urn value should be "4" and not a print statement here.
> 

Yes, I was surprised to see nothing, but when I briefly examined the code and executed the generated control file it looked like this:

```
import logging
logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '
                    'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')
import sys
sys.path.append('/tmp/utils')

def add_one(number):
    """A small decorated test function with extra nesting."""
    def do_nothing():
        pass
    do_nothing()
    return number + 1

result = add_one(3)
```

and the returned value is the cmd_output. So I assumed you are interested in the print statements instead and explicitely added the check for that to first cover the case and second to check whether that's the expected behaviour.

> > ```diff
> > +        # TODO: Check the control file output...
> > ```
> 
> Why did you remove the code checking the control file output above?
> 

Again, just PoC to see whether it's possible to turn remote execution into local for selftesting.

> > ```diff
> > +        # TODO: Remove the contorl file
> > ```
> 
> Not sure what this TODO means.
> 

Note that the final version should take care of the /tmp/tmp* files produced by this test

> > ```diff
> >      def test_run_remote_object(self):
> >          """Test that a remote object can be run properly."""
> > ```
> > 
> > 
> > As a bonus we'd get coverage for the `return` introduced in this series.
> 
> Perhaps we could find a way to test the return value by preserving the mocks, I will take a look.
> 

That would be another way, but without actually executing the control files it'd just be mocking. I think turning remote execution into local one should be safer and more accurate. Anyway I'm fine with either. A selftest for the returned value would be appreciated as it's not clear to me how it really works.

> > I'd really want you to consider adding those functional selftests as it can decrease the amount of mocking in the selftests.
> 
> What is bad about mocking? I understand these tests as unit tests and less as functional tests, the focus
> remaining on the tested utility and not taken too far away into the overall integration of such.
> 

Depends on the level of testing. Unit-test is great, but it'd be nice to check it all works together. Not essential, but nice to have.

> > If you decide not to do that please remove the additional mocks you added as they are not actually used and might give false impression that they are.
> 
> Can you be a bit more specific which additional mocks do you mean? All the mocks I have added that I know of are actually used
> so I am confused here.

The `@mock.patch('aexpect.remote_door.remote.remote_login', _local_login)` is not used at all (double-checked via debugger). But you are right that `@mock.patch('aexpect.remote_door.remote.copy_files_to', mock.MagicMock())` is needed.

pevogam:
Hi @ldoktor, I guess you took your example too literally and wasn't sure which parts you mean as something to implement and which were remnants from your own environment. To add more specific comments:

> > 
> > What is the reason for removing the `copy_files_to` from the class? It is used by most unit tests and
> > not just by the `test_run_remote_decorator`.
> 
> Again, it was interfering with the other mock and the fastest way to create a PoC was to remove it. Of course it'd have to be present for the other tests (unless we turn all of the tests into functional ones).

I guess by "interfering with the other mock" you mean that you wanted the real thing and not the mock? Because I have tested everything and the two mocks work fine together.

> > > ```diff
> > > @@ -88,6 +97,7 @@ class RemoteDoorTest(unittest.TestCase):
> > >                            "r'Wonderland is fun', wanderer=None)\n",
> > >                            control_lines)
> > >  
> > > +    @mock.patch('aexpect.remote_door.remote.copy_files_to', _local_copy)
> > >      def test_run_remote_decorator(self):
> > >          """Test that a remote decorated function can be run properly."""
> > >          @remote_door.run_remotely
> > > @@ -98,25 +108,15 @@ class RemoteDoorTest(unittest.TestCase):
> > >                  pass
> > >  
> > >              do_nothing()
> > > +            print("sample output")
> > > ```
> > 
> > 
> > What is the reason for the print statement? We are not checking logs or anything.
> 
> As you can see in the selftest, the "sample output" is the only thing that got returned when calling the `add_one` function. I also expected the `4` to be returned, but that did not happened. Perhaps I'm doing something wrong but running this selftest resulted in this behaviour.

I see. Interesting! The PR if fairly old and I first have to double check that the implementation of a return argument is complete (we haven't even put this feature to use yet).

> > > ```diff
> > > +        ret = add_one(session, 3)  # pylint: disable=E1121
> > > +        self.assertEqual(ret, "sample output\n")
> > > ```
> > 
> > 
> > The `ret`-urn value should be "4" and not a print statement here.
> 
> Yes, I was surprised to see nothing, but when I briefly examined the code and executed the generated control file it looked like this:
> 
> ```
> import logging
> logging.basicConfig(level=logging.DEBUG, format='%(module)-16.16s '
>                     'L%(lineno)-.4d %(levelname)-5.5s| %(message)s')
> import sys
> sys.path.append('/tmp/utils')
> 
> def add_one(number):
>     """A small decorated test function with extra nesting."""
>     def do_nothing():
>         pass
>     do_nothing()
>     return number + 1
> 
> result = add_one(3)
> ```
> 
> and the returned value is the cmd_output. So I assumed you are interested in the print statements instead and explicitely added the check for that to first cover the case and second to check whether that's the expected behaviour.

Alright, will check and come back to you on this. Thanks for the finding!

> > > ```diff
> > > +        # TODO: Remove the contorl file
> > > ```
> > 
> > 
> > Not sure what this TODO means.
> 
> Note that the final version should take care of the /tmp/tmp* files produced by this test

Noted.

> > > ```diff
> > >      def test_run_remote_object(self):
> > >          """Test that a remote object can be run properly."""
> > > ```
> > > 
> > > 
> > > As a bonus we'd get coverage for the `return` introduced in this series.
> > 
> > 
> > Perhaps we could find a way to test the return value by preserving the mocks, I will take a look.
> 
> That would be another way, but without actually executing the control files it'd just be mocking. I think turning remote execution into local one should be safer and more accurate. Anyway I'm fine with either. A selftest for the returned value would be appreciated as it's not clear to me how it really works.

+1

> > > I'd really want you to consider adding those functional selftests as it can decrease the amount of mocking in the selftests.
> > 
> > 
> > What is bad about mocking? I understand these tests as unit tests and less as functional tests, the focus
> > remaining on the tested utility and not taken too far away into the overall integration of such.
> 
> Depends on the level of testing. Unit-test is great, but it'd be nice to check it all works together. Not essential, but nice to have.

Alright.

> > > If you decide not to do that please remove the additional mocks you added as they are not actually used and might give false impression that they are.
> > 
> > 
> > Can you be a bit more specific which additional mocks do you mean? All the mocks I have added that I know of are actually used
> > so I am confused here.
> 
> The `@mock.patch('aexpect.remote_door.remote.remote_login', _local_login)` is not used at all (double-checked via debugger). But you are right that `@mock.patch('aexpect.remote_door.remote.copy_files_to', mock.MagicMock())` is needed.

The remote login is used as a call within:
```python
def share_remote_objects(session, control_path, host="localhost", port=9090,
                         os_type="windows", extra_params=None):
    """
    Create and share remote objects from a remote location over the network.
    ...
    """
    logging.info("Sharing the remote objects over the network")
    ...
    remote.copy_files_to(session.host, transfer_client,
                         session.username, session.password, transfer_port,
                         local_path, remote_path, timeout=10)
    middleware_session = remote.wait_for_login(session.client, session.host, session.port,
                                               session.username, session.password,
                                               session.prompt, session.linesep)
```

at least according to the API but I guess this part wasn't actually covered due to the mocked session. I guess extending to more functional testing could also help out making better use of the mock then. Will take a look too.

pevogam:
Hi @ldoktor, I have converted all unit tests to functional ones adding some more to improve the overall coverage. Somewhat not surprisingly, there are two complications arising out of this:

1) While all unit tests pass in my own environment some actually hand in the CI. Do you mind running them in your environment and telling me if you detect some troubles?

2) There are lots of warnings for unclosed files that I think I have seen before in aexpect but functional test conversion brings them to the unit test output stage. Things like `ResourceWarning: unclosed file` and `ResourceWarning: subprocess XXXX is still running`. I make sure to close all started sessions but to no avail.

All return argument functionality works as expected now and I added a few irrelevant fixes from the extended coverage.

tests/test_remote_door.py
```diff
@@ -12,118 +12,199 @@
...
+
+@mock.patch('aexpect.remote_door.remote.copy_files_to', _local_copy)
+@mock.patch('aexpect.remote_door.remote.wait_for_login', _local_login)
 class RemoteDoorTest(unittest.TestCase):
     """Unit test class for the remote door."""
 
     def setUp(self):
-        self.session = mock.MagicMock(name='session')
-        self.session.client = "ssh"
+        self.session = RemoteSession("sh", prompt="#", client="ssh")
```
ldoktor, line 47: The custom `prompt` here leads to infinity hang in Travis

tests/test_remote_door.py
```diff
@@ -12,118 +12,199 @@
...
+        os.mkdir(remote_door.REMOTE_PYTHON_PATH)
```
ldoktor, line 48: In case this directory already exists the setup fails (eg. on a terminated previous execution. How about `if os.path.isdir` or `os.path.exists`? (I don't want you to cover all cases, I'm just suggesting to be lenient towards the most common failures)

tests/test_remote_door.py
pevogam: Oh I see, let me comment below.

tests/test_remote_door.py
pevogam: Yes, I was wandering about the same but then realized that the directory should never be there if we want to execute unit tests and this is a good form of fast fail. Otherwise we are in the risk of writing into a preexisting directory and lose somebody's data (even not replacing the entire folder). But I also went through some inconvenience because of this so you have a point and I might make it more tolerant albeit also slightly more dangerous.

pevogam:
> Thank you, @pevogam, it seems to work well now. The selftests are hanging in the Travis CI as well as on my system due to a custom `prompt`. How about something like this?
> 
> ```diff
> diff --git a/tests/test_remote_door.py b/tests/test_remote_door.py
> index 6ff1222..2200de7 100644
> --- a/tests/test_remote_door.py
> +++ b/tests/test_remote_door.py
> @@ -44,8 +44,9 @@ class RemoteDoorTest(unittest.TestCase):
>      """Unit test class for the remote door."""
>  
>      def setUp(self):
> -        self.session = RemoteSession("sh", prompt="#", client="ssh")
> -        os.mkdir(remote_door.REMOTE_PYTHON_PATH)
> +        self.session = RemoteSession("sh", client="ssh")
> +        if not os.path.isdir(remote_door.REMOTE_PYTHON_PATH):
> +            os.mkdir(remote_door.REMOTE_PYTHON_PATH)
>  
>      def tearDown(self):
>          for control_file in glob.glob("tmp*.control"):
> ```

Strangely the container in my environment I tested this in had a very basic shell prompt before and this would not have worked there even though I tried it but it works well now in a different container. All tests passed on my side too and it seems that simply using the default prompt was the culprit of it all. Thanks for testing this!

I added this final change and pushed, everything passes now in the CI as well.

ldoktor (approved):
Thank you, looks useful and seems to work well now.

PR #68, opened 2020-09-26
Remove pre-release pylint issues

pevogam:

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -455,7 +455,7 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
     try:
         remote_object = Pyro4.Proxy("PYRONAME:" + object_name + "@" + host + ":" + str(port))
         remote_object._pyroBind()
-    except Pyro4.errors.PyroError:
+    except Pyro4.errors.PyroError as error:
         if not session:
             raise
 
@@ -471,7 +471,7 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
                 break
             time.sleep(1)
         else:
-            raise OSError("Local object sharing failed:\n%s" % output)
+            raise OSError("Local object sharing failed:\n%s" % output) from error
         logging.debug("Local object sharing output:\n%s", output)
         logging.getLogger("Pyro4").setLevel(10)
 
@@ -502,7 +502,7 @@ def get_remote_objects(session=None, host="localhost", port=0):
     try:
         remote_objects = flame.connect(host + ":" + str(port))
         remote_objects._pyroBind()
-    except Pyro4.errors.PyroError:
+    except Pyro4.errors.PyroError as error:
         if not session:
             raise
 
@@ -518,7 +518,7 @@ def get_remote_objects(session=None, host="localhost", port=0):
             time.sleep(1)
             control_log = session.cmd("cat " + REMOTE_CONTROL_LOG)
         else:
-            raise OSError("Local objects sharing failed:\n%s" % control_log)
+            raise OSError("Local objects sharing failed:\n%s" % control_log) from error
         logging.debug("Local objects sharing output:\n%s", control_log)
 
         remote_objects = flame.connect(host + ":" + str(port))
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -408,7 +408,7 @@ class DaemonLoop(threading.Thread):
         :param pyro_daemon: daemon for the remote python objects
         :type pyro_daemon: Pyro4.Daemon object
         """
-        threading.Thread.__init__(self)
+        super().__init__()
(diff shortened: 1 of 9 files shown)


aexpect/remote_door.py
```diff
@@ -747,12 +747,15 @@ def import_remote_exceptions(exceptions):
     # the errors modules and even extracting the exception classes from there
     class RemoteCustomException(Exception):
         """Standard class to instantiate during remote expection deserialization."""
+        def __init__(self):
+            super(RemoteCustomException, self).__init__()
+            self.__customclass__ = None
+
```
ldoktor, line 754: would be probably simpler to just use:

```diff
diff --git a/aexpect/remote_door.py b/aexpect/remote_door.py
index 5bd874c..83fbef4 100644
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -748,9 +748,7 @@ def import_remote_exceptions(exceptions):
     # the errors modules and even extracting the exception classes from there
     class RemoteCustomException(Exception):
         """Standard class to instantiate during remote expection deserialization."""
-        def __init__(self):
-            super().__init__()
-            self.__customclass__ = None
+        __customclass__ = None
 
     def recreate_exception(class_name, class_dict):
         logging.debug("Remote exception %s data: %s", class_name, class_dict)
```

aexpect/remote_door.py
```diff
@@ -471,7 +471,7 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
                 break
             time.sleep(1)
         else:
-            raise OSError("Local object sharing failed:\n%s" % output)
+            raise OSError("Local object sharing failed:\n%s" % output) from Pyro4.errors.PyroError
```
ldoktor, line 474: You need to put here the actual instance and not just the class to get the full picture.

aexpect/remote_door.py
```diff
@@ -518,7 +518,8 @@ def get_remote_objects(session=None, host="localhost", port=0):
             time.sleep(1)
             control_log = session.cmd("cat " + REMOTE_CONTROL_LOG)
         else:
-            raise OSError("Local objects sharing failed:\n%s" % control_log)
+            raise OSError("Local objects sharing failed:\n"
+                          "%s" % control_log) from Pyro4.errors.PyroError
```
ldoktor, line 522: The same here

aexpect/utils/process.py
```diff
@@ -28,7 +28,7 @@ class CmdError(Exception):
     """Error describing failed command execution"""
 
     def __init__(self, command=None, result=None):
-        super(CmdError, self).__init__()
+        super().__init__()
```
ldoktor, line 31: Nice but it'd be nice to replace the direct calls used elsewhere as well. Would you like to take care of them now, later or should I do that?

... I'm talking about these:

```diff
diff --git a/aexpect/client.py b/aexpect/client.py
index 4c3665b..9e1f5fc 100644
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -501,8 +501,8 @@ class Tail(Spawn):
         self._add_close_hook(Tail._close_log_file)
 
         # Init the superclass
-        Spawn.__init__(self, command, a_id, auto_close, echo, linesep,
-                       pass_fds, encoding)
+        super().__init__(command, a_id, auto_close, echo, linesep,
+                         pass_fds, encoding)
         if thread_name is None:
             self.thread_name = "tail_thread_%s_%s" % (self.a_id,
                                                       str(command)[:10])
```

ldoktor (changes requested):
Thank you, I was about to fix these but luckily I looked at notifications first. You nailed most of them, the only reason for change is the `raise from`. The other comments are just notes for your consideration but the `raise from` needs to be changed.

aexpect/remote_door.py
pevogam: :+1:

aexpect/utils/process.py
pevogam: I guess I have missed some, I was scouting for all of them but my IDE seems to have skipped some nevertheless. Sorry about that, will double check and fix this one. In case I miss more don't hesitate to ping me.

aexpect/remote_door.py
pevogam: I guess I had a mental blackout here due to the mechanic changes of all other instances, thanks for spotting! The benefits of having more than one pair of eyes on the same changes ^^

pevogam:
Hi @ldoktor, I committed all the changes you requested and the CI is passing but sometimes (like it just happened in one variant here) has a race condition in one of the unit tests that I have fixed in #67. Feel free to take a look at #67 so that we can merge it and fix the bit of randomness that made it in the CI.

ldoktor (approved):
Thank you, looks good. I'll take a look at the #67, I'll try to squeeze it into my schedule today but not promising anything.

PR #70, opened 2020-10-23
Make two ssh options changeable and support VMware Vcenter login

xiaodwan:
1) Make 'PreferredAuthentications' and UserKnownHostsFile changeable.
2) VMware vCenter login support

Signed-off-by: Xiaodai Wang <xiaodwan@redhat.com>

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -258,7 +258,9 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
                  r"[Cc]onnection.*closed", r"[Cc]onnection.*refused",
                  r"[Pp]lease wait", r"[Ww]arning", r"[Ee]nter.*username",
                  r"[Ee]nter.*password", r"[Cc]onnection timed out", prompt,
-                 r"Escape character is.*"],
+                 r"Escape character is.*",
+                 r"Command>",
+                 r"[Ll]ost connection"],
                 timeout=timeout, internal_timeout=0.5)
             output += text
             if match == 0:  # "Are you sure you want to continue connecting"
@@ -288,17 +290,19 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
                     else:
                         msg = "Got username prompt after password prompt"
                     raise LoginAuthenticationError(msg, text)
-            elif match == 5:  # "Connection closed"
+            elif match == 5:   # "Connection closed"
                 raise LoginError("Client said 'connection closed'", text)
-            elif match == 6:  # "Connection refused"
+            elif match == 6:   # "Connection refused"
                 raise LoginError("Client said 'connection refused'", text)
             elif match == 11:  # Connection timeout
                 raise LoginError("Client said 'connection timeout'", text)
-            elif match == 7:  # "Please wait"
+            elif match == 15:  # Connection lost
+                raise LoginError("Disconnected to vm on remote host", text)
+            elif match == 7:   # "Please wait"
                 if debug:
                     logging.debug("Got 'Please wait'")
                 timeout = 30
-            elif match == 8:  # "Warning added RSA"
+            elif match == 8:   # "Warning added RSA"
                 if debug:
                     logging.debug("Got 'Warning added RSA to known host list")
             elif match == 12:  # prompt
@@ -308,6 +312,17 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
             elif match == 13:  # console prompt
                 logging.debug("Got console prompt, send return to show login")
                 session.sendline()
+            elif match == 14:  # VMware vCenter command prompt
+                # Some old vsphere version (e.x. 6.0.0) needs to enable first.
+                cmd = 'shell.set --enabled True'
+                logging.debug(
+                    "Got VMware VCenter prompt, "
+                    "send '%s' to enable shell first", cmd)
+                session.sendline(cmd)
+                logging.debug(
+                    "Got VMware VCenter prompt, "
+                    "send 'shell' to launch bash")
+                session.sendline('shell')
         except ExpectTimeoutError as error:
             # sometimes, linux kernel print some message to console
             # the message maybe impact match login pattern, so send
@@ -327,7 +342,9 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
 def remote_login(client, host, port, username, password, prompt, linesep="\n",
                  log_filename=None, log_function=None, timeout=10,
                  interface=None, identity_file=None,
-                 status_test_command="echo $?", verbose=False, bind_ip=None):
+                 status_test_command="echo $?", verbose=False, bind_ip=None,
+                 preferred_authenticaton='password',
+                 user_known_hosts_file='/dev/null'):
     """
     Log into a remote host (guest) using SSH/Telnet/Netcat.
 
@@ -353,6 +370,8 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
             cmd_status_output() and friends).
     :param bind_ip: ssh through specific interface on
                     client(specify interface ip)
+    :param preferred_authenticaton: Given value of PreferredAuthentications in ssh
+    :param user_known_hosts_file: Given value of UserKnownHostsFile in ssh
     :raise LoginError: If using ipv6 linklocal but not assign a interface that
                        the neighbour attache
     :raise LoginBadClientError: If an unknown client is requested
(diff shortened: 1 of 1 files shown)


xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.convert_vm_to_ovirt.esx.vm.7_0.linux.latest6.arch_x86_64.raw_f.NFS.rhv_upload.rhv_direct.rhv_noverifypeer.nopreallocated.it_vddk: PASS (657.74 s)
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.AV_WIN7.esx_67.rhev.rhv_upload.it_vddk.esx_uri: PASS (923.14 s)

xiaodwan:
This PR comes from those three PRs which support v2v testing.
https://github.com/avocado-framework/avocado-vt/pull/2046
https://github.com/avocado-framework/avocado-vt/pull/2038
https://github.com/avocado-framework/avocado-vt/pull/2327

xiaodwan:
@pevogam Please help to review this PR, thanks.

pevogam:
Hi @xiaodwan, first of all thanks for this! With regard to the migrated options and since you are on it, do you think could could include the ones from https://github.com/avocado-framework/avocado-vt/pull/2720 like "Lost connection" as well? In this way we could restore the full previous functionality with just one round.

xiaodwan:
@pevogam I'll check that.

xiaodwan:
@pevogam Please help to check it, thanks.

pevogam:
@xiaodwan, I will take a look early next week and do some testing, then come back to you. It looks good at a first glance.

xiaodwan:
@pevogam Thanks, This patch is a MUST NEED for v2v testing when latest avocado-vt is used.

aexpect/remote.py
```diff
@@ -288,17 +290,19 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
                     else:
                         msg = "Got username prompt after password prompt"
                     raise LoginAuthenticationError(msg, text)
-            elif match == 5:  # "Connection closed"
+            elif match == 5:   # "Connection closed"
```
pevogam, line 293: I see some of these comments having three spaces before the inline comment while others have two, I am not sure if this is intentional.

aexpect/remote.py
```diff
@@ -308,6 +312,15 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
             elif match == 13:  # console prompt
                 logging.debug("Got console prompt, send return to show login")
                 session.sendline()
+            elif match == 14:  # VMware vCenter command prompt
+                # Some old vsphere version (e.x. 6.0.0) needs to enable first.
+                cmd = 'shell.set --enabled True'
+                logging.debug(
+                    "Got VMware VCenter prompt, send '%s' to enable shell first", cmd)
```
pevogam, line 319: I would break this into something more readable like
```python
logging.debug("Got VMware VCenter prompt, "
              "send '%s' to enable shell first", cmd)
```

aexpect/remote.py
```diff
@@ -308,6 +312,15 @@ def handle_prompts(session, username, password, prompt=PROMPT_LINUX,
...
+                session.sendline(cmd)
+                logging.debug(
+                    "Got VMware VCenter prompt, send 'shell' to launch bash")
```
pevogam, line 322: This one too.

aexpect/remote.py
```diff
@@ -353,6 +368,8 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
             cmd_status_output() and friends).
     :param bind_ip: ssh through specific interface on
                     client(specify interface ip)
+    :param preferred_authenticaton: The preferred authentication of SSH connection
+    :param user_known_hosts_file: one or more files to use for the user host key database
```
pevogam, line 372: I would either have both parameters described with capital letters or both with small letters but just commenting here.

aexpect/remote.py
```diff
@@ -403,14 +420,16 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
 
 def wait_for_login(client, host, port, username, password, prompt,
                    linesep="\n", log_filename=None, log_function=None,
-                   timeout=240, internal_timeout=10, interface=None):
+                   timeout=240, internal_timeout=10, interface=None,
+                   preferred_authenticaton='password'):
```
pevogam, line 424: Could we also add the `user_known_hosts_file` in the end here, making it available for waiting login calls? This shouldn't harm current API and adds all options to all useful locations.

aexpect/remote.py
```diff
@@ -425,15 +444,17 @@ def wait_for_login(client, host, port, username, password, prompt,
         try:
             return remote_login(client, host, port, username, password, prompt,
                                 linesep, log_filename, log_function,
-                                internal_timeout, interface, verbose=verbose)
+                                internal_timeout, interface, verbose=verbose,
+                                preferred_authenticaton=preferred_authenticaton)
```
pevogam, line 448: We could then also call this the current `user_known_hosts_file` argument.

aexpect/remote.py
```diff
@@ -425,15 +444,17 @@ def wait_for_login(client, host, port, username, password, prompt,
...
         except LoginError as error:
             logging.debug(error)
             verbose = True
         time.sleep(2)
     # Timeout expired; try one more time but don't catch exceptions
     return remote_login(client, host, port, username, password, prompt,
                         linesep, log_filename, log_function,
-                        internal_timeout, interface)
+                        internal_timeout, interface,
+                        preferred_authenticaton=preferred_authenticaton)
```
pevogam, line 457: We could then also call this the current `user_known_hosts_file` argument.

pevogam (commented):
Hi @xiaodwan, I ran some testing and it works well, just left some minor comments if you think we should make use of them.

aexpect/remote.py
xiaodwan: I kept the change in original avocado-vt PR. 
I think it's for keeping them at same column to look tidy?
Anyway, I googled the PEP8, it only says 'Inline comments should be separated by at least two spaces from the statement.' IMO, it's fine. But I could change it back if you think it's a style in aexpect.
Thanks.

aexpect/remote.py
xiaodwan: Yes, agree.

aexpect/remote.py
xiaodwan: Yes, agree.

aexpect/remote.py
xiaodwan: Sure, absolutely agree.

aexpect/remote.py
xiaodwan: Sure.

aexpect/remote.py
xiaodwan: Sure

aexpect/remote.py
xiaodwan: I updated it to ':param user_known_hosts_file: Given value of UserKnownHostsFile in ssh', is it ok?

xiaodwan:
(1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.AV_WIN7.esx_67.rhev.rhv_upload.it_vddk.esx_uri: PASS (922.21 s)

xiaodwan:
Hi @pevogam, I updated the PR as your option and tested it. Please help to review it again.
BTW, I don't like plain password is printed in log by 'logging.debug("Got password prompt, sending '%s'", password)'. we could hide some characters by '*', what do you think?

aexpect/remote.py
pevogam: No particular style and I thought the same thing about at least two spaces. I was only wondering because some of the other inline comments used two spaces (the minimum) and these used three even though they are neighboring lines. But this is just a minor thing and not too important, let's keep the VT formatting for now and unify later.

aexpect/remote.py
pevogam: Sure, either one works at least for a minor consistency - I checked your change and it is fine.

pevogam:
> Hi @pevogam, I updated the PR as your option and tested it. Please help to review it again.
> BTW, I don't like plain password is printed in log by 'logging.debug("Got password prompt, sending '%s'", password)'. we could hide some characters by '*', what do you think?

You make a good point. I would usually argue that this is useful for debugging where aexpect is only used for tests but this is far from the truth and aexpect could be used for other purposes as well. OTOH anyone using it for a production level access would probably rely on SSH keys anyway and avoid the trouble of logging the password altogether. Also anyone using the password method will need to store it somewhere for the API calls as well and currently aexpect doesn't provide any additional security for passing a password around be it for linux or windows connections. Password obfuscation coupled with some additional password handling could generally be the subject of a future PR.

pevogam (approved):
LGTM, @ldoktor do you think you could help review this PR or recommend somebody interested in doing this?

aexpect/remote.py
ldoktor: Well ideally a commit should either change style, refactor or implement new features. It should not change multiple things to simplify reviews as well as reverts, but let's keep it for now.

ldoktor (approved):
Hello @xiaodwan, it makese sense but next time please split your patches so they include either style changes, refatctors or new features and only one feature per commit (in this example I'd rather see 2 commits with preferred_auth/user_known_hosts and second with the VMware VCenter prompt optionally followed by unifying style commit. Also it'd be nice to include link (commit or PR) to the source this is backported from.

Anyway we do want to get all of those backported and I don't expect reverts so merging it as is to speedup things. Mentioning this for the next time because even backport should follow the best practices.

aexpect/remote.py
lmr: We should use `python-black` to mass format it and then all the discussions become moot.

lmr (commented):
I agree with @ldoktor, we should separate style and functional changes.

And as I mentioned, we should probably use `python-black` accross the board to make style discussions moot.

PR #74, opened 2021-02-03
support proxy server

YongxueHong:
Signed-off-by: Yongxue Hong <yhong@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -27,6 +27,7 @@ import select
 import subprocess
 import locale
 import logging
+import random
 
 from aexpect.exceptions import ExpectError
 from aexpect.exceptions import ExpectProcessTerminatedError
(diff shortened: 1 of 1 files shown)


ldoktor (changes requested):
Hello @YongxueHong, I'd like to know more about this feature you are proposing, because it's quite intrusive change which results into many similar branches, lacks documentation and the rpc server part is also missing.

Would you please share more about your usage and the reasons for this patch so we can evaluate whether it's something that should belong to Aexpect or whether another solution would fit better?

Regards,
Lukáš

YongxueHong:
Hi @ldoktor 
It's just a draft, just want to find somewhere to record it.
I will close this PR.
Sorry for the inconvenience.

ldoktor:
Hello @YongxueHong, it'd be still nice to see the details. Perhaps an alternative solution already exists or someone might suggest another way of solving this. Or you might interest other users that would help you testing and improving your solution...

PR #75, opened 2021-03-02
Remove vestigial python shebang line from rss_client.py

mmathesius:
The file `aexpect/rss_client.py` contains an unnecessary `#!/usr/bin/python` line that, when packaged, results in the following error when `rpmlint` is run:

```
python3-aexpect.noarch: E: non-executable-script /usr/lib/python3.9/site-packages/aexpect/rss_client.py 644 /usr/bin/python 
```

This PR simply removes that extraneous line to clean up the error.

### DIFF
--- a/aexpect/rss_client.py
+++ b/aexpect/rss_client.py
@@ -1,5 +1,3 @@
-#!/usr/bin/python
-
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation; either version 2 of the License, or


mmathesius:
Note: this `rpmlint` error was identified during the Fedora package review to bring back the non-modular version of the `python-aexpect` package. https://bugzilla.redhat.com/show_bug.cgi?id=1930954

pevogam (commented):
Hi @mmathesius,

This sounds reasonable, however the `rss_client.py` can indeed be executed as a standalone script as you can see from the `__main__` check at the end of the module. How about simply fixing the file permissions of the script within the git tree instead?

mmathesius:
Hi @pevogam, thanks for the feedback.

> This sounds reasonable, however the `rss_client.py` can indeed be executed as a standalone script as you can see from the `__main__` check at the end of the module.

Yes. And even after this change to satisfy `rpmlint`, it can be executed as a standalone script by running

```
python /usr/lib/python3.x/site-packages/aexpect/rss_client.py ...
```

> How about simply fixing the file permissions of the script within the git tree instead?

This PR has already changed the file permissions within the git tree from 100755 → 100644. But that doesn't change anything since the installation process was already stripping the execute permissions from the module code being installed into ` /usr/lib/python3.x/site-packages/aexpect/` since directly executable scripts shouldn't normally be going into the location.

pevogam (approved):

pevogam:
@lmr @ldoktor @clebergnu Do you mind taking a look at the Travis CI for this PR? It changed names from travis.org so I suspect the settings of this project have to be updated. As the CI contains quite a picky lint checkers it would be good to run them on this PR just in case.

willianrampazzo (approved):

ldoktor (approved):
Well this is tough. It is mainly used as a library, but can be used in clients as well. But since we don't usually rely on the shebang and it's more-of-a-library let's remove it and rely on proper execution in case people know what they are doing.

PR #78, opened 2021-05-27
Not stop reading output when there's no new output from any attempt

chloerh:
Sometimes output could be printed out a little slow and there could
be no new output from an attempt. But that doesn't mean there's not
going to be new output just seconds later. If we break the loop
every time there's no new output, we could end up with raising
`ExpectError: Unknown error occurred`. This commit should fix the
issue.

Signed-off-by: Haijiao Zhao <haizhao@redhat.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -855,7 +855,7 @@ class Expect(Tail):
             data = self.read_nonblocking(internal_timeout,
                                          end_time - time.time())
             if not data:
-                break
+                continue
             # Print it if necessary
             if print_func:
                 for line in data.splitlines():


aexpect/client.py
```diff
@@ -855,7 +855,7 @@ def read_until_output_matches(self, patterns, filter_func=lambda x: x,
             data = self.read_nonblocking(internal_timeout,
                                          end_time - time.time())
             if not data:
-                break
+                continue
```
ldoktor, line 858: Well when the poller returns `poll_status` there have to be something to be read. If there is anything to be read, the `read_nonblocking` should return it and we should proceed, otherwise we should finish, shouldn't we?

ldoktor: Hmm, you are right, when the data is ignored (we do `decode(self.encoding, 'ignore')`) therefore it's possible that we receive something but there is nothing returned. Let me do another set of tests...

ldoktor: Yes, tested that with a modified `read_nonblocking` and in case we receive only characters that are ignored by encoding it returns immediately. This fix addresses the `read_until_output_matches` part but even the `read_nonblocking` should improve. Anyway the `read_nonblocking` fix won't address this `read_until_output_matches` issue as we are polling here as well so let me merge this and suggest the `read_nonblocking` fix in a separate PR.

Thank you for noticing this, it must have been really hard to find.

ldoktor (changes requested):
Oh actually this implementation could lead to infinite loop in case the pipe is closed as we always continue without checking the end-time condition as even 0s poll returns a success when the pipe is closed. Would you please change the `continue` for:

```python
if not data:
    if time.time > end_time:
        break
    else:
        continue
```

that would be the safest solution which would only potentially suffer a finite busy loop. A less secure but IMO safe-enough solution would be to improve `read_nonblocking` to report number of read characters as well as the output. Let me propose that in a followup PR so we can decide which would be more useful.

ldoktor:
@chloerh would you please take a look at the proposed solution in #79? The benefit of that solution would be proper detection of the pipeline close, potential cons is that it might not detect all cases, but I'm most certain it should (unless we expect the pipeline to be reconnected in the future, in such case this solution with the additional time-check should be better).

Please let me know which solution you prefer.

chloerh:
@ldoktor Thanks for taking look at this pr. I agree that the current pr could `potentially suffer a finite busy loop`, so I'm happy to close this one since #79 is there to fix the problem. 
I have run my case with #79 and it also fixes the issue. Thanks again :handshake:

PR #79, opened 2021-05-31
Make sure read_until_output_matches waits for all data

ldoktor:
In case the output produces non-decodable characters the `read_until_output_matches` can return earlier as it considers that a closed pipeline. Let's detect the closed pipeline by counting the raw bytes rather than decoded value with ignored values.

This is an alternative solution of the issue described in: https://github.com/avocado-framework/aexpect/pull/78

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -866,10 +866,12 @@ class Expect(Tail):
             if not poll_status:
                 raise ExpectTimeoutError(patterns, output)
             # Read data from child
-            data = self.read_nonblocking(internal_timeout,
-                                         end_time - time.time())
-            if not data:
+            read, data = self._read_nonblocking(internal_timeout,
+                                                end_time - time.time())
+            if not read:
                 break
+            if not data:
+                continue
             # Print it if necessary
             if print_func:
                 for line in data.splitlines():
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -153,6 +153,7 @@ class Spawn(object):
         assert os.path.isdir(base_dir)
 
         self.command = command
+        self._aexpect_helper = None
 
         # Remember some attributes
         self.auto_close = auto_close
@@ -178,12 +179,13 @@ class Spawn(object):
         # Start the server (which runs the command)
         if command:
             helper_cmd = utils_path.find_command('aexpect_helper')
-            sub = subprocess.Popen([helper_cmd],
-                                   shell=True,
-                                   stdin=subprocess.PIPE,
-                                   stdout=subprocess.PIPE,
-                                   stderr=subprocess.STDOUT,
-                                   pass_fds=pass_fds)
+            self._aexpect_helper = subprocess.Popen([helper_cmd],
+                                                    shell=True,
+                                                    stdin=subprocess.PIPE,
+                                                    stdout=subprocess.PIPE,
+                                                    stderr=subprocess.STDOUT,
+                                                    pass_fds=pass_fds)
+            sub = self._aexpect_helper
             # Send parameters to the server
             sub.stdin.write(("%s\n" % self.a_id).encode(self.encoding))
             sub.stdin.write(("%s\n" % echo).encode(self.encoding))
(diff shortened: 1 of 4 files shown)


chloerh:
I have run some tests with this pr:

**Before** this fix:
```
JOB ID     : a9dbcca5060a72d2f3de14b334adaa7a7e3dbb99
JOB LOG    : /root/avocado/job-results/job-2021-05-31T03.52-a9dbcca/job.log
 (1/1) type_specific.io-github-autotest-libvirt.pcibridge.positive_test.vm_with_pcie_br.no_br: ERROR: Unknown error occurred while looking for patterns ['[Aa]re you sure', '[Pp]assword:\\s*', '\\(or (press|type) Control-D to continue\\):\\s*$', '[Gg]ive.*[Ll]ogin:\\s*$', '(?<![Ll]ast )[Ll]ogin:\\s*$', '[Cc]onnection.*closed', '[Cc]onnection.*refused', '[P... (28.37 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 28.76 s
```

After
```
JOB ID     : a9b1772ee6eee8d6a82576e71deb79e3b7ad85c1
JOB LOG    : /root/avocado/job-results/job-2021-05-31T03.58-a9b1772/job.log
 (1/1) type_specific.io-github-autotest-libvirt.pcibridge.positive_test.vm_with_pcie_br.no_br: PASS (41.52 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 41.88 s
```
 I have also run **another 24 test cases** of our most common test set. No new failure. 

Thank you for fixing this issue for us !

chloerh (approved):

aexpect/client.py
```diff
@@ -753,21 +762,34 @@ def read_nonblocking(self, internal_timeout=None, timeout=None):
...
-                return data
+                return read, data
             if poll_status:
-                new_data = os.read(expect_pipe, 1024).decode(self.encoding,
-                                                             "ignore")
-                if not new_data:
-                    return data
-                data += new_data
+                raw_data = os.read(expect_pipe, 1024)
+                if not raw_data:
```
chunfuwen, line 773: Could the line put behind line 776, where raw_data had been decoded ?

aexpect/client.py
```diff
@@ -753,21 +762,34 @@ def read_nonblocking(self, internal_timeout=None, timeout=None):
...
+                    return read, data
+                read += len(raw_data)
```
chunfuwen, line 775: Could this line put  just behind raw_data = os.read(expect_pipe, 1024) ?

chunfuwen, line 775: it sound like read  need be calculated based on  return value of  len(raw_data.decode(self.encoding, "ignore")), otherwise read is always consistent with raw_data.

aexpect/client.py
```diff
@@ -852,10 +874,12 @@ def read_until_output_matches(self, patterns, filter_func=lambda x: x,
             if not poll_status:
                 raise ExpectTimeoutError(patterns, output)
             # Read data from child
-            data = self.read_nonblocking(internal_timeout,
-                                         end_time - time.time())
-            if not data:
+            read, data = self._read_nonblocking(internal_timeout,
+                                                end_time - time.time())
+            if not read:
```
chunfuwen, line 879: it sounds like this line need put behind line 882 if this PR wants to satisfy requirement in https://github.com/avocado-framework/aexpect/pull/78.

ldoktor: It should be the opposite way around, `read` is the length of the raw bytes read while the `data` is the decoded (with ignore) string.

Therefore when the pipe is closed, both `read` and `data` will be false (`0` and `''`). When the read is solely composed of ignore chars, the `read` will be non-zero (amount of read bytes, at least `1`) but `data` will be empty (`''`).

ldoktor: `read >= len(data)`

chunfuwen: thank for explanation

chunfuwen (approved):
lgtm

clebergnu (approved):
LGTM.  Avocado has also dropped Python 3.5, so this puts them in sync.

PR #80, opened 2021-06-28
Spec file cleanup resulting from downstream package review.

mmathesius:
The Fedora `python-aexpect` package was recently brought back as a non-modular package. To un-retire the package, it underwent a fresh Fedora package review process which resulted in cleanup to the SPEC file. (See https://bugzilla.redhat.com/show_bug.cgi?id=1930954 for discussion.)

This PR brings the downstream package SPEC file cleanup to the upstream SPEC file.

### DIFF
--- a/python-aexpect.spec
+++ b/python-aexpect.spec
@@ -1,5 +1,3 @@
-%global srcname aexpect
-
 # Conditional for release vs. snapshot builds. Set to 1 for release build.
 %if ! 0%{?rel_build:1}
     %global rel_build 1
@@ -7,7 +5,7 @@
 
 # Settings used for build from snapshots.
 %if 0%{?rel_build}
-%global gittar		%{srcname}-%{version}.tar.gz
+%global gittar		aexpect-%{version}.tar.gz
 %else
 %if ! 0%{?commit:1}
 %global commit		9f53e380ef4f4eb3cdf04845deab27be5d3ed7ac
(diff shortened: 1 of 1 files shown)


python-aexpect.spec
```diff
@@ -1,13 +1,11 @@
-%global srcname aexpect
-
```
ldoktor, line 2: well the srcname makes sense to me, why removing and hard coding it everywhere?

python-aexpect.spec
```diff
@@ -17,29 +15,27 @@
...
-Name: python-%{srcname}
+Name: python-aexpect
 Version: 1.6.2
-Release: 1%{?gitrel}%{?dist}
+Release: 2%{?gitrel}%{?dist}
 Summary: Aexpect is a python library to control interactive applications
-Group: Development/Tools
 
-License: GPLv2
+License: GPLv2+
```
ldoktor, line 29: This is actually more accurate as our license in code is GPLv2+

python-aexpect.spec
```diff
@@ -17,29 +15,27 @@
...
 URL: https://github.com/avocado-framework/aexpect
 
 %if 0%{?rel_build}
-Source0: https://github.com/avocado-framework/%{srcname}/archive/%{version}.tar.gz#/%{gittar}
+Source0: %{url}/archive/%{version}/%{gittar}
```
ldoktor, line 33: Interesting, are the variables case insensitive? (I can only see `URL` being defined, not `url`)

ldoktor (approved):
Seems fine and works well athough would you please clarify the importance of the "srcname" removal? Anyway I guess Fedora packagers know more than me about the right format and rpmlint is not complaining either.

@clebergnu would you have a look at this? (not sure whether you're still pushing custom rpms to our repo)

python-aexpect.spec
mmathesius: @ldoktor Thanks for reviewing.

I made this change based on a suggestion during the Fedora package review process for related package `python-avocado`: https://bugzilla.redhat.com/show_bug.cgi?id=1930948:

> Opinionated suggestion: The spec file somewhat arbitrarily(?) uses "avacado", "%{srcname}" and "%{pkgname}". I'd stick to "avocado" for better readability, I don't expect the name of the project to be updated so often to justify this complexity.

python-aexpect.spec
mmathesius: The tag names (eg., `URL`, `Name`, `Version`, etc.) _are_ case-insensitive, but macro names are case-sensitive. The built-in `%{url}` macro is indeed lower case, and it takes its value from whatever case you happen to use for `URL`.

lmr (approved):
Looks good. @ldoktor already brought up some small clarification questions that reflect the one doubt I have about the removal of `srcname`.

clebergnu (approved):
LGTM, RPMs build nicely as expected.

PR #81, opened 2021-07-01
Add six dependency to setup.py

ldoktor:
Aexpect requires six, let's include it in our setup.py.

Signed-off-by: Lukáš Doktor <ldoktor@redhat.com>

PR #82, opened 2021-07-01
Switch CI from travis-ci to github actions

ldoktor:
travis-ci requires periodic renewals, let's switch to github actions instead.

Note I modified the matrix to avoid over-using the good will of github and simply went with the eldest and latest python versions as that should hopefully cover most of the issues.

Also I included #81 as without it the builder is failing.


ldoktor:
Here is the CI result: https://github.com/ldoktor/aexpect/actions/runs/989839580 (I had to switch from python-3.5 to 3.6 because 3.5 does not contain the R1732 but it was the previous configuration anyway...

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+  push:
+  workflow_dispatch:
+
+jobs:
+  selftests:
+    name: ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.x]
```
ana, line 15: I'd make last version explicit and write directly 3.9. 
Travis was testing in 3.6, 3.7 and 3.8. Is there any reason to skip 3.6 and 3.7 here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: pip install -r requirements-travis.txt
+      - run: make check
```
ana, line 23: You can put this two actions in the same run command. Or instead adding a name to the step so it's easier to read in the web, like:

```
- name: installing requeriments
  run: pip install -r requirements-travis.txt
- name: run checks
  run: make check
```

.github/workflows/ci.yml
ldoktor: We always want to cover the latest python so I'd prefer to follow the latest one of the supported pythons. As for the matrix I don't want to unnecessary abuse the github free workers. Covering old(ish) and latest python should catch most of the issues.

.github/workflows/ci.yml
ldoktor: I know but these one-liners are self-explanatory enough. If there were many steps than bundling them together would make sense but putting a name to "pip install" or "make check" seems unnecessary to me. Anyway I can change that if you think otherwise...

.github/workflows/ci.yml
ldoktor: btw, Ana, thank you for your example ci.yml, I simply copied it and adjusted :-)

ana: @ldoktor you're welcome!
The tests are very small so I don't think it'll be abuse of builders. And if all these python versions are supported it'd be a good idea to test all of them. 
About 3.x, in some months 3.10 will be released and the actions will switch automatically, this might trigger some false regressions...

.github/workflows/ci.yml
ana: Current lines are correct, this is a matter of taste.

ana:
This looks good to me, only missing testing all the supported python versions.

lmr (approved):
LGTM

.github/workflows/ci.yml
lmr: I like the idea of checking it in 3.7, you never know in which distributions people are running aexpect on.

lmr (approved):
LGTM

setup.py
```diff
@@ -28,4 +28,5 @@
           packages=['aexpect',
                     'aexpect.utils'],
           scripts=['scripts/aexpect_helper'],
-          test_suite='tests')
+          test_suite='tests',
+          install_requires=['six'])
```
clebergnu, line 32: This is correct.  But given the current Python 3 only status, we must drop six ASAP.  I've opened https://github.com/avocado-framework/aexpect/issues/83.

clebergnu (approved):
LGTM, thanks!

PR #87, opened 2021-09-24
Pylint style fixes

ldoktor:
Addressing the newly added C0209 and W1514 checks failures.

### DIFF
--- a/scripts/aexpect_helper
+++ b/scripts/aexpect_helper
@@ -37,9 +37,9 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
     a_id = sys.stdin.readline().strip()
     echo = sys.stdin.readline().strip() == "True"
     readers = sys.stdin.readline().strip().split(",")
-    command = sys.stdin.readline().strip() + " && echo %s > /dev/null" % a_id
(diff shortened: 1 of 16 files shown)


scripts/aexpect_helper
```diff
@@ -119,11 +118,11 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
             # Write shell PID to file
             server_log.info('Writing shell PID file %s', shell_pid_filename)
-            with open(shell_pid_filename, "w") as file_obj:
+            with open(shell_pid_filename, "w", encoding="utf-8") as file_obj:
```
lmr, line 121: Ah, an option for open that I didn't know about. Living and learning.

lmr (approved):
Looks good, approved!

scripts/aexpect_helper
ldoktor: Well in py2 one had to use codecs to open with encoding, in py3 they added it as an argument and in 3.10 they added `encoding="locale"` which should be what we use in avocado (`locale.getpreferredencoding()`)

pevogam:
Great job @ldoktor, I was so behind with GitHub notifications that I couldn't see this on time. In any case thanks for this!

Issue #84, opened 2021-07-09
Sometimes cmd_output_safe() can't get all output


huangyum:
Sometimes, the command output is not captured by cmd_output_safe().  It's because some part of the command output is accidentally discarded during the polling routine: 

```
# @read_until_output_matches()
...
        output = ""
        end_time = time.time() + timeout
        while True:
            try:
                poll_timeout_ms = max(0, (end_time - time.time()) * 1000)
                poll_status = poller.poll(poll_timeout_ms)
            except select.error:
                break
            if not poll_status:
                raise ExpectTimeoutError(patterns, output)         # <- if it happens at the 2nd or behind round, the `output` collected by the 1st round won't return to the caller
            # Read data from child
            data = self.read_nonblocking(internal_timeout,
                                         end_time - time.time())
            if not data:
                break
            # Print it if necessary
            if print_func:
                for line in data.splitlines():
                    print_func(line)
            # Look for patterns
            output += data
...

# @cmd_output_safe()
...
        while (time.time() - start_time) < timeout:
            try:
                out += self.read_up_to_prompt(0.5)
                success = True
                break
            except ExpectTimeoutError as error:                    # <- then we'll never see the missing output at here
                self.sendline()
...
```

pevogam:
@ldoktor Have you experienced anything like this? Or should this issue be closed if we can't reproduce it?

ldoktor:
Thank you, @huangyum, for the nice description and please take a look at the #88 which should address this issue. Let me mention the simple reproducer here:

```pyhton
import aexpect
a=aexpect.ShellSession("sh")
q=a.cmd_output_safe("echo A; sleep 2; echo B", timeout=100000)
print(q)
```

PR #88, opened 2021-10-04
Include previous output in cmd_output_safe

ldoktor:
The ShellSession.cmd_output_safe uses 0.5 inner timeout to get the
output which is not stored for following iterations. Similarly the
exceptions only use the current iteration output, leaving the previous
iteration(s) output discarded.

As this test is time-sensitive (or long) this commit introduces a
conditional selftest which is only executed when AEXPECT_TIME_SENSITIVE
is set. This should be performed at least before a release.

Reported-by: Yumei Huang yuhuang@redhat.com

Fixes: #84

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ COMMIT_DATE=$(shell git log --pretty='format:%cd' --date='format:%Y%m%d' -n 1)
 MOCK_CONFIG=default
 
 all:
-	@echo "make check - Runs tree static check, unittests and functional tests"
+	@echo "make check - Runs tree static check, unittests and functional tests. Some tests are only executed when AEXPECT_TIME_SENSITIVE=yes is set."
 	@echo "make clean - Get rid of scratch and byte files"
 	@echo "make source - Create source package"
 	@echo "make install - Install on local system"
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -14,6 +14,7 @@
 
 # selftests pylint: disable=C0111,C0111
 
+import os
 import random
 import string
 import sys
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ COMMIT_DATE=$(shell git log --pretty='format:%cd' --date='format:%Y%m%d' -n 1)
 MOCK_CONFIG=default
 
 all:
-	@echo "make check - Runs tree static check, unittests and functional tests"
+	@echo "make check - Runs tree static check, unittests and functional tests. Some tests are only executed when AEXPECT_TIME_SENSITIVE=yes is set."
```
pevogam, line 13: I assume this is the best location for this detail and there is no document or any other location that could also be updated.

tests/test_client.py
```diff
@@ -112,6 +113,21 @@ def test_cmd_timeout(self):
                                   f"({type(details)}) was raised "
                                   f"using command {cmd} ({self.cmds})")
 
+    @unittest.skipUnless(os.environ.get('AEXPECT_TIME_SENSITIVE'),
+                         "AEXPECT_TIME_SENSITIVE env variable not set")
+    def test_cmd_output_with_inner_timeout(self):
+        """
+        cmd_output_safe uses 0.5s inner timeout, make sure all lines are
```
pevogam, line 120: Perhaps it is better to just write "Make sure all lines are present in the output." in the main docstring, adding the `cmd_output_safe` subsentence below as a detail and complete separate sentence.

pevogam (approved):
LGTM, thanks for the quickly cooked up PR about this!

Makefile
ldoktor: Yeah, I was looking for a documentation and found out there is none :-/

tests/test_client.py
ldoktor: Well this is cmd_output_safe-specific test, the read_nonblocking is not usually executed in a loop.

pevogam: Could that for instance rather be mentioned in `test_cmd_output_safe_with_inner_timeout`?

clebergnu (approved):
LGTM, thanks!

PR #89, opened 2021-11-11
Add logging namespace for all relevant aexpect modules

pevogam:
As Aexpect is a library used by other applications, it makes sense
to limits its logging and not log directly to the root logger.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -54,6 +54,8 @@ from aexpect.utils import wait as utils_wait
 
 _THREAD_KILL_REQUESTED = threading.Event()
 
+LOG = logging.getLogger(__name__)
+
 
 def kill_tail_threads():
     """
@@ -1168,7 +1170,7 @@ class ShellSession(Expect):
         """
         if safe:
             return self.cmd_output_safe(cmd, timeout)
-        logging.debug("Sending command: %s", cmd)
+        LOG.debug("Sending command: %s", cmd)
         self.read_nonblocking(0, timeout)
         self.sendline(cmd)
         try:
@@ -1206,7 +1208,7 @@ class ShellSession(Expect):
                 terminates while waiting for output
         :raise ShellError: Raised if an unknown error occurs
         """
-        logging.debug("Sending command (safe): %s", cmd)
+        LOG.debug("Sending command (safe): %s", cmd)
         self.read_nonblocking(0, timeout)
         self.sendline(cmd)
         out = ""
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -55,6 +55,8 @@ from aexpect.exceptions import ExpectTimeoutError
 from aexpect.exceptions import ExpectProcessTerminatedError
 from aexpect import rss_client
 
+LOG = logging.getLogger(__name__)
+
 #: prompt to be used for shell sessions on linux machines (default)
 PROMPT_LINUX = r"^\[.*\][\#\$]\s*$"
 #: prompt to be used for shell sessions on windows machines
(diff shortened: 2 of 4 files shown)


pevogam:
@beraldoleal This PR is allowing us to restore Aexpect's logging in Avocado, any review in your spare cycles is appreciated. I haven't tested it yet and will go about that right now.

pevogam:
Noting down here, it might be better to just go all the way to using `log` on each message in order to avoid a collision where Avocado VT might do `from aexpect.remote import *` and end up reimporting the logging module. Of course, we can also provide a public API through `__all__` module attribute and importing wildcards are a bad idea to begin with.

aexpect/client.py
```diff
@@ -54,6 +54,8 @@
 
 _THREAD_KILL_REQUESTED = threading.Event()
 
+logging = log.getLogger(__name__)
```
willianrampazzo, line 57: Most of the places updated recently on the Avocado side use `LOG` (uppercase) as the variable name. There are still old entries in the utilities using lowercase, so it is not uniform yet. It would be good to make this change uniform within and across the projects.

Anyway, this is just a comment, not a blocker.

pevogam: Yes, I am also wondering about the `s/logging/log/` change. In any case I have mostly seen projects use `log` and the capitalized `LOG` seems to be an Avocado thing. Granted, this is an Avocado subproject but it could also be generalized beyond this scope I think. In any case not important, I am fine by either choice.

willianrampazzo: > In any case not important, I am fine by either choice.

Same here.

pevogam:
Since there aren't more opinions from the last weeks and the project size is fairly small I took the liberty to bring this PR to a more finalized state and did the replacement not just for imports but also for all emitted logging messages. Any further reviews that could bring this PR to fruition are welcome.

aexpect/client.py
```diff
@@ -54,6 +54,8 @@
 
 _THREAD_KILL_REQUESTED = threading.Event()
 
+log = logging.getLogger(__name__)
```
beraldoleal, line 57: My only suggestion here is to rename this constant to LOG instead of log to follow pep8.

beraldoleal (commented):
Hi @pevogam this LGTM, just one small comment.

aexpect/client.py
pevogam: Thanks @beraldoleal for the review. Can you point me to the PEP8 guideline about this? I think the pushed branched passed multiple checks, including PEP8 ones. Also, we discussed the same option with @willianrampazzo in https://github.com/avocado-framework/aexpect/pull/89#discussion_r747521840 and settled on the fact that this is optional. Finally, I already edited all all messages so I am somewhat unwilling to go through the entire thing again but I understand that this might be preferred for `log` to be deemed a constant, just wondering how solid this potential PEP8 guideline is.

beraldoleal: https://www.python.org/dev/peps/pep-0008/#constants

It is just a suggestion, not a blocker.

pevogam: No worries, I will go through it one more time with some IDE replacements. I rather meant redoing this as problematic because on my side it also involves backpropagating patches to other fork branches as well as RPMs. As there was no further input from last time with Willian I thought the only justification for using `LOG` is compatibility with Avocado projects and decided to go with a different final change anyways. The PEP8 provides further justification that I didn't have at the time.

willianrampazzo (approved):

beraldoleal (approved):

PR #90, opened 2021-12-06
Fix remote_login with nc client for windows guest

xiaodwan:
In commit 23829fca09f75badf651acb593e7d4e4e2a98001, a new parameter
was introduced, but it broke the nc client. This patch fixes the
problem.

Signed-off-by: Xiaodai Wang <xiaodwan@redhat.com>

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -371,6 +371,8 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     :param status_test_command: Command to be used for getting the last
             exit status of commands run inside the shell (used by
             cmd_status_output() and friends).
+    :param verbose: If True, log some stats using LOG.debug (RSS only), only
+            valid when client is ssh or nc.
     :param bind_ip: ssh through specific interface on
                     client(specify interface ip)
     :param preferred_authenticaton: Given value of PreferredAuthentications in ssh
@@ -405,7 +407,7 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     elif client == "telnet":
         cmd = f"telnet -l {username} {host} {port}"
     elif client == "nc":
-        cmd = f"nc {verbose} {host} {port}"
+        cmd = f"nc {' '.join(extra_params)} {host} {port}"
     else:
         raise LoginBadClientError(client)
 


xiaodwan:
The test result:
 (1/1) type_specific.io-github-autotest-libvirt.function_test_esx.positive_test.windows.uefi.win2019.esx_67.rhev.rhv_upload.it_vddk.vpx_uri: PASS (535.50 s)

xiaodwan:
Looks like the CI has problem.
@clebergnu @lmr @chunfuwen Could you help to review this PR? Because the 1.6.3 blocks v2v testing against windows guest, I'd like a new release of aexpect to fix the issue. Thanks.

xiaodwan:
+ @ldoktor

Yingshun (approved):
LGTM. The nc command was mistaken and tried to connect 'False'. And this fix works well in my env.

chunfuwen:
It sounds like previous fix( In commit 23829fc) miss the change in case:client == "nc"

chunfuwen (approved):
LGTM

dzhengfy (approved):
LGTM

dzhengfy:
@ldoktor , this fix is to correct the regression in https://github.com/avocado-framework/aexpect/pull/86 and important for whole v2v test. Could you help review and merge it if no concern? And a new aexpect tag is also needed. Thanks.

aexpect/remote.py
```diff
@@ -405,7 +404,7 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     elif client == "telnet":
         cmd = f"telnet -l {username} {host} {port}"
     elif client == "nc":
-        cmd = f"nc {verbose} {host} {port}"
+        cmd = f"nc {' '.join(extra_params)} {host} {port}"
```
beraldoleal, line 407: extra_params is used on 'ssh' and 'nc' now. However is not being used when using 'telnet'. If that is by design, then we should make it clear on the docstring.

aexpect/remote.py
```diff
@@ -383,8 +383,7 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     :return: A RemoteSession object.
     """
     extra_params = []
-    if verbose:
-        extra_params.append("-vv")
+    extra_params.append(f"{'-vv' if verbose else ''}")
```
beraldoleal, line 386: IIUC, this change is adding an empty string to the list in case of normal run (verbose=false). While this is valid, I find odd having a empty value inside a "list of extra parameters". Unless I'm missing something, the fix itself is on the line 407. And it will work even with this current logic here. So, is there any reason for you adding this empty string here?

beraldoleal: Also, I know that is not related to your change, but if you dont mind, could please, take this opportunity to add the 'verbose' argument to this method docstring, since it is missing?

beraldoleal (changes requested):

aexpect/remote.py
xiaodwan: Sure

aexpect/remote.py
xiaodwan: Yes, The change in L386 is not necessary.

aexpect/remote.py
xiaodwan: @beraldoleal I updated the docstring, please have a look, thanks.

clebergnu:
Closing / reopening to trigger CI.

clebergnu (approved):
LGTM.

beraldoleal (approved):

PR #97, opened 2022-01-12
client: Make sure aexpect_helper file descriptors are closed

bgartzi:
The commit fixes the bug listed below. It also fixes a file leak that was being reported when users were opening and closing sessions while keeping a reference to them even if they were closed.
The commit introduces the needed changes to make sure the file leak is fixed.

ID: 2003990

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -277,6 +277,25 @@ class Spawn:
             except OSError:
                 pass
 
+    def _close_aexpect_helper(self):
+        """
+        Makes sure the subprocess holding the aexpect_helper is terminated
+        and that the file descriptors related to it are closed. Then releases
+        the popen instance.
+        """
+        if self._aexpect_helper:
+            try:
+                self._aexpect_helper.wait(10)
+            except subprocess.TimeoutExpired:
+                self._aexpect_helper.kill()
+            finally:
+                if self._aexpect_helper.stdout:
+                    self._aexpect_helper.stdout.close()
+                if self._aexpect_helper.stdin:
+                    self._aexpect_helper.stdin.close()
+                self._aexpect_helper.wait(10)
+            self._aexpect_helper = None
+
     def get_id(self):
         """
         Return the instance's a_id attribute, which may be used to access the
@@ -374,13 +393,8 @@ class Spawn:
             # Remove all used files
             if 'AEXPECT_DEBUG' not in os.environ:
                 shutil.rmtree(os.path.join(BASE_DIR, f'aexpect_{self.a_id}'))
+            self._close_aexpect_helper()
             self.closed = True
-        if self._aexpect_helper:
-            try:
-                self._aexpect_helper.wait(10)
-            except subprocess.TimeoutExpired:
-                self._aexpect_helper.kill()
-                self._aexpect_helper.wait(10)
 
     def set_linesep(self, linesep):
         """
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -128,6 +128,30 @@ class CommandsTests(unittest.TestCase):
         self.assertIn("SECOND LINE", out)
         self.assertIn("THIRD LINE", out)
 
+    def test_fd_leak(self):
+        """
+        Check file descriptors are not being leaked
+        """
+        def get_proc_fds():
+            """
+            Returns a set containing the fd names opened under the process
+
+            :returns: set
+            """
+            # Ommitting the last one since it is the one opened to
+            # get the result from running the listdir method
+            process_fds = os.listdir(f"/proc/{os.getpid()}/fd")[:-1]
+            return set(process_fds)
+
+        fds_before = get_proc_fds()
+        session = client.ShellSession("sh")
+        session.close()
+        fds_after = get_proc_fds()
+        self.assertEqual(fds_after, fds_before,
+                         msg="fd leak: Closing the session didn't close "
+                             "the file descriptors")
+        session = None
+
 
 if __name__ == '__main__':
     unittest.main()


bgartzi:
@luckyh could you please have a check on this pull request?

qingwangrh:
python3 ConfigTest.py --testcase=block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset --platform=x86_64   --machines=q35 --guestname=Win10,Win2019 --nrepeat=3
(1/8) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: PASS (700.69 s)
 (2/8) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads.q35: PASS (665.84 s)
 (3/8) repeat1.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (267.45 s)
 (4/8) repeat1.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (299.42 s)
 (5/8) repeat2.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (251.43 s)
 (6/8) repeat2.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (229.32 s)
 (7/8) repeat3.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (252.03 s)
 (8/8) repeat3.Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.block_during_io.iozone_stress.system_disk.send_qmp.with_system_reset.q35: PASS (230.39 s)

LGTM

aexpect/client.py
```diff
@@ -277,6 +277,24 @@ def _close_reader_fds(self):
...
+                self._aexpect_helper.stdout.close()
+            try:
+                if self._aexpect_helper.stdin:
+                    self._aexpect_helper.stdin.close()
+            finally:
+                try:
+                    self._aexpect_helper.wait(10)
+                except subprocess.TimeoutExpired:
+                    self._aexpect_helper.kill()
+                    self._aexpect_helper.wait(10)
```
luckyh, line 296: To copy the exact way that `subprocess.call` handled this, I'd suggest calling `wait`/`kill` at first:

```suggestion
            try:
                try:
                    self._aexpect_helper.wait(10)
                except subprocess.TimeoutExpired:
                    self._aexpect_helper.kill()
            finally:
                if self._aexpect_helper.stdout:
                    self._aexpect_helper.stdout.close()
                try:
                    if self._aexpect_helper.stdin:
                        self._aexpect_helper.stdin.close()
                finally:
                    self._aexpect_helper.wait(10)
```


https://github.com/python/cpython/blob/be578e0c063dad1dbb273f86d5bc77e4e6f14583/Lib/subprocess.py#L345-L351
https://github.com/python/cpython/blob/be578e0c063dad1dbb273f86d5bc77e4e6f14583/Lib/subprocess.py#L1033-L1059

luckyh, line 296: And, at the end we could consider to release the `Popen` instances.
```python
self._aexpect_helper = None
```

bgartzi: I'll redo the commit right now, thank you

bgartzi: Yes, it looks way nicer like this, thank you

bgartzi:
Hi @luckyh,

I checked if the fact that different methods for getting the fds opened by a process open new file descriptors affected the suggested test.

In general, if there is not a difference in the file descriptors opened between both checks, the id-number of the file descriptor opened to do the check will be the same at first and at the end. Thus, we won't be facing the problems I was thinking about at first.

In fact, I made a quick check. I ran the test 100 times over the fixed code and it passed all the times. I also ran the test as many times as before over the previous commit and it failed 100 times. I checked the output of the sets and set differences and it seems that it makes sense.

I'm providing an updated commit including the test case in a few minutes.

Regards,

luckyh:
> Hi @luckyh,
> 
> I checked if the fact that different methods for getting the fds opened by a process open new file descriptors affected the suggested test.
> 
> In general, if there is not a difference in the file descriptors opened between both checks, the id-number of the file descriptor opened to do the check will be the same at first and at the end. Thus, we won't be facing the problems I was thinking about at first.
> 
> In fact, I made a quick check. I ran the test 100 times over the fixed code and it passed all the times. I also ran the test as many times as before over the previous commit and it failed 100 times. I checked the output of the sets and set differences and it seems that it makes sense.
> 
> I'm providing an updated commit including the test case in a few minutes.
> 
> Regards,

@bgartzi good to know that, thanks!

luckyh (approved):
ACK

luckyh:
Hi @ldoktor would you mind to help review this PR? The issue might be something related to https://github.com/avocado-framework/aexpect/pull/79.

aexpect/client.py
```diff
@@ -277,6 +277,28 @@ def _close_reader_fds(self):
...
+                    self._aexpect_helper.kill()
+            finally:
+                if self._aexpect_helper.stdout:
+                    self._aexpect_helper.stdout.close()
+                try:
+                    if self._aexpect_helper.stdin:
+                        self._aexpect_helper.stdin.close()
+                finally:
+                    self._aexpect_helper.wait(10)
+                self._aexpect_helper = None
```
ldoktor, line 300: This cleanup looks a bit unnecessary complicated to me, how about simply:

```python
if self._aexpect_helper:
    try:
        self._aexpect_helper.wait(10)
    except subprocess.TimeoutExpired:
        self._aexpect_helper.kill()
    finally:
        if self._aexpect_helper.stdout:
            self._aexpect_helper.stdout.close()
        if self._aexpect_helper.stdin:
            self._aexpect_helper.stdin.close()
        self._aexpect_helper.wait(10)
    self._aexpect_helper = None
```

I mean the second try/finally only catches stdin handling and not the stdout handling.

ldoktor (commented):
Thanks for this catch, indeed when the `wait` timeouts the stdin/stdout are not closed. Please consider the necessity of the second try/finally and either let me know why stdout does not require one and stdin does or unify both handling.

bgartzi:
Hi Lukáš,

Sorry for the late reply. We were considering the second try/catch as it's done that way in the Popen's `__exit__` method [0] since "Flushing a BufferedWriter may raise an error".

However, I'm starting to think that waiting and then cleaning by `self._aexpect_helper = None` might also do the work in a much simpler way. I mean, I consider that this way python should be calling the default clean up method from Popen and getting the things done by itself, am I wrong?

Thank you!

[0] https://github.com/python/cpython/blob/be578e0c063dad1dbb273f86d5bc77e4e6f14583/Lib/subprocess.py#L1033-L1059

aexpect/client.py
bgartzi: Hi,

As I commented before, considering your feedback and getting code simplified, even something like 
```python
    def _close_aexpect_helper(self):
        """
        Makes sure the subprocess holding the aexpect_helper is terminated
        and that the file descriptors related to it are closed. Then releases
        the popen instance.
        """
        if self._aexpect_helper:
            try:
                self._aexpect_helper.wait(10)
            except subprocess.TimeoutExpired:
                self._aexpect_helper.kill()
                self._aexpect_helper.wait(10)
            finally:
                self._aexpect_helper = None
```

Should have the work done by itself, as the problem was that saving an indirect reference to the process wasn't letting the python garbage collector clean the `Popen` related to the `self._aexpect_helper`. However, this way we wouldn't be manually assuring that the fds are closed, but relying on the Popen's cleanup routines instead.

According to some testing I'm running right now, even the now suggested code piece is closing the opened fds. Deciding whether one or the other fits better on the code should be related to wanting to get things done _manually_ or just relying on the `Popen` cleanup routines implemented on `subprocess.py`.

Please, @ldoktor, @luckyh let me know what do you think about this. I'll update the commit considering the feedback.

Thank you,
Beñat

luckyh: @bgartzi yes, I had realized this thing when I was going to give the suggestion https://github.com/avocado-framework/aexpect/pull/97#discussion_r783689453. But finally, for making the behavior predictable, I decided to not let the cleanup procedure rely on Python's GC. Or otherwise, I'm not sure if we might deal with some weird problem related to that in the future.

Thanks,

ldoktor:
> Hi Lukáš,
> 
> Sorry for the late reply. We were considering the second try/catch as it's done that way in the Popen's `__exit__` method [0] since "Flushing a BufferedWriter may raise an error".
> 
> However, I'm starting to think that waiting and then cleaning by `self._aexpect_helper = None` might also do the work in a much simpler way. I mean, I consider that this way python should be calling the default clean up method from Popen and getting the things done by itself, am I wrong?

Well that's what I thought as well but thanks to you I learned the pipes from `Popen.stdin|stdout|stderr` are not being cleaned by `Popen`'s `__del__`. Anyway the current version seems complete and does closes even unkillable processes. Thank you for this fix.

> 
> Thank you!
> 
> [0] https://github.com/python/cpython/blob/be578e0c063dad1dbb273f86d5bc77e4e6f14583/Lib/subprocess.py#L1033-L1059

ldoktor (approved):
Yep, looks good to me.

ldoktor:
As there are no further comments and one ack, let me merge this.

PR #98, opened 2022-09-24
Add session ops utility with some facilitator functions (v2)

pevogam:
Add a small library of frequent operations performed in aexpect sessions.

v1: Initial pull request, later split into manageable parts.
v2: Main session ops module contribution and selection of reusable session functions, rebased on most recent aexpect tag.

### DIFF
--- a/aexpect/ops.py
+++ b/aexpect/ops.py
@@ -1,0 +1,475 @@
+# This Python file uses the following encoding: utf-8
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Intra2net AG and aexpect contributors
+
+"""
+A list of frequent operations performed through remote sessions.
+
+Most commands passed to a session are routine invocations of common userland
+tools. Nevertheless some of them require a certain effort to set up because
+they are squeezed through the eye of the needle, the shell session.
+
+The functionality in this module assumes a remote host with the ordinary userland
+tools available. Some of it might just work on Windows remote hosts if the proper
+environment (i. e. Cygwin) is present.
+
+This module aims to provide convenient wrappers for butter-and-bread
+functionality on a Linux system. Currently, the implemented functions fall
+under a few categories: ``stat``, ``test``, common file ops, and utilities.
+
+stat:
+Allows passing an optional format argument (``-c``). Custom wrappers for
+extracting *atime*, *ctime*, and *mtime*, etc exist.
+
+test:
+Allows for testing existence, permissions and other properties of files.
+
+file ops:
+Simple file operations executed through sessions. Most functions boil down to
+executing simple commands like ls, tar, md5sum, etc. If these operations return
+non-0, a :py:class:`RuntimeError` is raised, containing the commands's error
+message. All functions :py:func:`shlex.quote` their args for better security.
+
+utilities:
+More complex linux utilities for tarball extraction, file hashing, etc.
+
+..warning:: This module runs commands on raw shell sessions and although we usually
+    quote the main arguments, we do not perform complex checks for escapes. It is
+    meant only for some convenience in simple enough use cases and using it might
+    result in data loss or worse.
+"""
+
+import logging
+from shlex import quote
+
+from aexpect.exceptions import ShellCmdError
+# Need this import for sphinx and other documentation to produce links later on
+# from .client import ShellSession
+
+LOG = logging.getLogger(__name__)
+
+
+###############################################################################
+# stat(2)
+###############################################################################
+
+
+def stat(session, path, fmt=None):
+    """
+    Wrapper for ``stat``.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :param str fmt: optional format passed to ``-d``
+    :returns: standard output of stat(1) on ``path``
+    :rtype: str
+    """
+    if fmt is None:
+        return session.cmd_output(f"stat {quote(path)}")
+
+    if not isinstance(fmt, str) or fmt[0] != "%":
+        raise RuntimeError(f"{fmt} is not a valid format string for stat(1)")
+
+    return session.cmd_output(f"stat -c {fmt} {quote(path)}")
+
+
+def get_atime(session, path, human_readable=False):
+    """
+    Query the access time.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :param bool human_readable: whether to use human readable format or seconds
+                                since Epoch
+    :returns: standard output of stat(1) querying the access time of ``path``
+    :rtype: str
+    """
+    return stat(session, path, fmt=r"%x" if human_readable else "%X")
+
+
+def get_mtime(session, path, human_readable=False):
+    """
+    Query the modification time.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :param bool human_readable: whether to use human readable format or seconds
+                                since Epoch
+    :returns: standard output of stat(1) querying the last modification time
+              of ``path``
+    :rtype: str
+    """
+    return stat(session, path, fmt=r"%z" if human_readable else r"%Z")
+
+
+def get_ctime(session, path, human_readable=False):
+    """
+    Query the change time.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :param bool human_readable: whether to use human readable format or seconds
+                                since Epoch
+    :returns: standard output of stat(1) querying the last change time of
+              ``path``
+    :rtype: str
+    """
+    return stat(session, path, fmt=r"%y" if human_readable else r"%Y")
+
+
+def get_size(session, path):
+    """
+    Query the file size.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :returns: standard output of stat(1) querying the size of ``path``
+    :rtype: int
+    """
+    return int(stat(session, path, fmt=r"%s"))
+
+
+###############################################################################
+# test(1)
+###############################################################################
+
+
+def test(session, path, flags):
+    """
+    Wrapper for ``test``.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to test
+    :param str flags: flags passed to test
+    :returns: status of test(1) on ``path`` as True or False
+    :rtype: bool
+    """
+    return session.cmd_status(f"test {flags} {quote(path)}") == 0
+
+
+def is_directory(session, path):
+    """
+    Check if a directory exists on a given remote host.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the directory to check
+    :returns: whether the directory exists
+    :rtype: bool
+    """
+    return test(session, path, flags="-d")
+
+
+def is_regular_file(session, path):
+    """
+    Check if a regular file exists on a given remote host.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the regular file to check
+    :returns: whether the regular file exists
+    :rtype: bool
+    """
+    return test(session, path, flags="-f")
+
+
+def is_nonzero_size_file(session, path):
+    """
+    Check if a regular file exists on a given remote host.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the regular file to check
+    :returns: whether the regular file exists and is nonzero size
+    :rtype: bool
+    """
+    return test(session, path, flags="-s")
+
+
+###############################################################################
+# file ops
+###############################################################################
+
+
+def ls(session, path, quote_path=True, flags="-1UNq"):  # pylint: disable=C0103
+    """
+    Run `ls` in given directory through a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: absolute or relative path to list
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to ls command
+    :returns: names of files in dir (without path components)
+    :rtype: [str]
+
+    Just like :py:func:`os.listdir`, does not include file names starting with
+    dot (`'.'`)
+    """
+    cmd = f'ls {flags} {quote(path)}' if quote_path else f'ls {flags} {path}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to ls {path}: {output}')
+    return output.splitlines()
+
+
+def move(session, source, target, quote_path=True, flags=""):
+    """
+    Move file or directory from source to target / rename source to target.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str source: full path of existing file or directory
+    :param str target: full path of new file or directory
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to mv command
+    :raises: :py:class:`RuntimeError` if mv returns non-zero exit status
+
+    Calls `mv source target` through a session. See `man mv` for what source
+    and target can be and what behavior to expect.
+    """
+    cmd = f'mv {flags} {quote(source)} {quote(target)}' if quote_path \
+        else f'mv {flags} {source} {target}'
+    status, output = session.cmd_status_output(cmd)
+    if status != 0:
+        raise RuntimeError(f"Failed to move {source} to {target}: "
+                           f"{output.strip() if output else ''}")
+
+
+def copy(session, source, target, quote_path=True, flags=""):
+    """
+    Copy file or directory from source to target.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str source: full path of existing file or directory
+    :param str target: full path of new file or directory
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to cp command
+    :raises: :py:class:`RuntimeError` if cp returns non-zero exit status
+
+    Calls `cp source target` through a session. See `man cp` for what source
+    and target can be and what behavior to expect.
+    """
+    cmd = f'cp {flags} {quote(source)} {quote(target)}' if quote_path \
+        else f'cp {flags} {source} {target}'
+    status, output = session.cmd_status_output(cmd)
+    if status != 0:
+        raise RuntimeError(f"Failed to copy {source} to {target}: "
+                           f"{output.strip() if output else ''}")
+
+
+def remove(session, path, quote_path=True, flags="-fr"):
+    """
+    Remove a directory including its contents.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: directory to remove
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to rm command
+    :raises: :py:class:`RuntimeError` if deletion fails
+
+    Calls `rm -rf`.
+    """
+    cmd = f'rm {flags} {quote(path)}' if quote_path else f'rm {flags} {path}'
+    status, output = session.cmd_status_output(cmd)
+    if status != 0:
+        raise RuntimeError(f"Failed to remove {path}: "
+                           f"{output.strip() if output else ''}")
+
+
+def make_tempdir(session, template=None):
+    """
+    Create a temporary directory in a safe way through a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param template: directory name template as used in the standard command
+    :type template: str or None
+    :returns: name of temporary directory
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if mktemp returns non-zero exit status
+
+    Calls `mktemp -d`, refer to `man` for more info.
+    """
+    cmd = f'mktemp -d {template}' if template is not None else 'mktemp -d'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to create temporary directory: {output}')
+    return output
+
+
+def make_tempfile(session, template=None):
+    """
+    Create a temporary file in a safe way through a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param template: file name template as used in the standard command
+    :type template: str or None
+    :returns: name of temporary file
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if mktemp returns non-zero exit status
+
+    Calls `mktemp`, refer to `man` for more info.
+    """
+    cmd = f'mktemp {template}' if template is not None else 'mktemp'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to create temporary file: {output}')
+    return output
+
+
+def cat(session, filename, quote_path=True, flags=""):
+    """
+    Get contents of a text file from a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str filename: full path of file
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to cat command
+    :returns: file contents
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if cat command fails
+
+    Should only be used for very short files without tabs or other fancy
+    contents. Otherwise better download file or use some other method.
+    """
+    cmd = f'cat {flags} {quote(filename)}' if quote_path else f'cat {flags} {filename}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to cat {filename}: {output}')
+    return output
+
+
+def sed_replace(session, pattern, replacement, filename):
+    """
+    Replace a pattern in a file using sed, raising an error if nothing is found.
+
+    :param session: session of the remote host
+    :type session: :py:class:`RemoteSession`
+    :param str pattern: pattern to replace
+    :param str replacement: string to replace the pattern for
+    :param str filename: file with content to be replaced
+    :raises: :py:class:`AssertionError` if nothing was replaced
+
+    Using sed is convenient, but it does not error out when nothing is replaced
+    and this can lead to false-positives when used inside tests.
+    """
+    LOG.info("Replacing {pattern} with {replacement} in {filename}")
+    # for lines matching pattern, replace and print them (the p command won't work)
+    cmd = f"""sed -ri '/{pattern}/{{
+s//{replacement}/g
+w /dev/stdout
+}}' {filename}"""
+    output = session.cmd(cmd)
+    # make sure we have at least one line replaced
+    if len(output.splitlines()) == 0:
+        raise RuntimeError(f"No matches found by sed for pattern {pattern}")
+
+
+def grep(session, expr, path, check=False, flags="-a"):
+    """
+    Invoke ``grep`` remotely searching for an expression in a path.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str expr: search expression
+    :param str path: file to search
+    :param bool check: whether to quietly run grep for a boolean check
+    :param str flags: extra flags passed to ``grep`` on the command line
+    :returns: whether there is a match or not or what ``grep`` emits on stdout
+              if the check mode is disabled
+    :rtype: bool or str
+    :raises: ShellCmdError if the check mode is disabled and status is nonzero
+    """
+    grep_command = f"grep {flags} '{expr}' {quote(path)}"
+    status, output = session.cmd_status_output(grep_command)
+    if check:
+        return status == 0
+    if status != 0:
+        raise ShellCmdError(grep_command, status, output)
+    return output
+
+
+###############################################################################
+# utilities
+###############################################################################
+
+
+def hash_file(session, filename, method='md5'):
+    """
+    Calculate hash of given file from a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str filename: full path of file that should be hashed
+    :param str method: method for hashing
+    :returns: hash of file, a hex number
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if hash command failed (e.g. file not
+             found) or resulting output does not have expected format
+    """
+    cmd = f"{method}sum"
+
+    # run cmd on shell
+    status, output = session.cmd_status_output(cmd + ' ' + quote(filename))
+    if output:
+        output = output.strip()
+    if status != 0 or not output:
+        raise RuntimeError(f'Could not hash {filename} using {cmd}: {output}')
+
+    # parse output
+    hash_str = output.split(maxsplit=1)[0].lower()
+
+    # check that all chars are hex
+    if hash_str.strip('0123456789abcdef'):
+        raise RuntimeError('Resulting hash string has unexpected characters: '
+                           + hash_str)
+    return hash_str
+
+
+def extract_tarball(session, tarball, target_dir, flags="-ap"):
+    """
+    Extract tarball to given dir from a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str tarball: full path of tar file that should be extracted
+    :param str target_dir: name of directory where tarball should be extracted
+    :param str flags: extra flags passed to ``tar`` on the command line
+    :raises: :py:class:`RuntimeError` if tar command returned non-null
+    """
+    cmd = f'tar -C {quote(target_dir)} {flags} -xf {quote(tarball)}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to extract {tarball} to {target_dir}: {output}')


pevogam:
The CI problem detected here is not related to these changes. Not sure how it wasn't detected in earlier pull requests but I will push one to fix it next.

pevogam:
All CI problems fixed and PR can now be fully reviewed.

ldoktor:
> Hello @pevogam, I gave it another try after this long time and my feeling is still the same. Although it simplifies certain workflows I don't think it fits aexpect library. I can imagine it as another library that would generate command strings that would be then consumed by aexpect (or other library that allows shell-like interface).

By this I mean something like:

```python
import aexpect
from aexpect_ops import *

session = aexpect.ShellSession("sh")
session.cmd(stat("/tmp/a"))
session.cmd(copy("/tmp/a", "/tmp/b"))
...
```

pevogam:
Hi @ldoktor,

> Hello @pevogam, I gave it another try after this long time and my feeling is still the same. Although it simplifies certain workflows I don't think it fits aexpect library. I can imagine it as another library that would generate command strings that would be then consumed by aexpect (or other library that allows shell-like interface).

You might be able to conclude that from the current selection of commands but there are commands that would require only `sendline`, some would require sending special characters like ^C or ^D, others would require varied timeouts and arguments to the `session.cmd` calls or different `session_cmd*` calls. So the current one-string-preparation-for-session-command approach would reduce a few of the simplest commands but will not generalize over future ones.

> My main problem is the size, the higher-level aim with security checks that are present in many places but in another they are missing (eg. `stat(a, "/tmp/a", fmt="%; echo this could as well be formatting your hard drive; ")`.

For sure it might have some limitations that we all expect to be pointed out during review and patched up before merging. I think you could notice that this pull request combines lots of different code from different authors in our team and previously added in different times when they were actually needed. So some unification is definitely to ask but I don't think will be hard to accomplish and this is what reviews are for - to discover maintainers' minimum uniform interface requirements (and leave some divergence that will be unavoidable aside).

> I can create a repo for you and make you an admin of that, if you are interested, but I don't like the idea of having this in the core aexpect (note ideally I'd extract and publish the `remote.py` separately as well, but it's already there...)

Why would the remote be considered separate from aexpect? The entire point of our previous arguments in the past was that it makes more sense in aexpect than in Avocado VT and I think if extends the `client` module quite nicely adding the file transfer and session login/prompts support. The same can now be said about the remote door module which extends this further with remote function execution and remote object method calls. But letting all of this aside, do you think that a separate project would make things easier? One thing we seem to have learned from the past is that fragmenting projects into too many small ones adds a lot of deployment, versioning, and maintenance overhead resulting in hours of packaging, debugging and testing of version compatibility and so on.

> Anyway @clebergnu, @lmr please take a look at this and if you decide it belongs here, feel free to give it a try.

Another point I want to add here is that we find lots of contributors on the Avocado and Avocado VT side reimplementing a lot of these functions on top of a session again and again which confirms the need to have single reusable source that everybody could instead simply spend their time extending and saving the rest of their costs. As these utilities depend only on aexpect and all they require is a session I don't think it makes sense to keep them in Avocado or Avocado VT repos specializing in something else.

Regarding a separate aexpect repo, can you elaborate more on what are the benefits from keeping current aexpect to the minimal code base and functionality it currently has? I would assume reduced maintenance cost and contributor activity is one of them but how would this compare to the cost of managing a separate repo and integrating the two mentioned above? The aexpect repo has not been very active in the last years but this could be viewed both as something good and something not so good (example of the second being contributors that did not want to use it because they thought the project has been abandoned at least in their replies to me). So all in all I would be glad to hear your reasoning regarding keeping an already small project like aexpect bare bone and core-only since at least in my opinion it would make sense to split into multiple repos projects that are much larger than this so that the costs from the splitting are outweighed by the benefits of separate maintenance.

ldoktor:
> Hi @ldoktor,
> 
> > Hello @pevogam, I gave it another try after this long time and my feeling is still the same. Although it simplifies certain workflows I don't think it fits aexpect library. I can imagine it as another library that would generate command strings that would be then consumed by aexpect (or other library that allows shell-like interface).
> 
> You might be able to conclude that from the current selection of commands but there are commands that would require only `sendline`, some would require sending special characters like ^C or ^D, others would require varied timeouts and arguments to the `session.cmd` calls or different `session.cmd_*` calls. So the current one-string-preparation-for-session-command approach would reduce a few of the simplest commands but will not generalize over future ones.
> 

I haven't seen such commands in this PR and therefore I mentioned that solution. For that kind of usage one would indeed had to depend directly on session usage.

> > My main problem is the size, the higher-level aim with security checks that are present in many places but in another they are missing (eg. `stat(a, "/tmp/a", fmt="%; echo this could as well be formatting your hard drive; ")`.
> 
> For sure it might have some limitations that we all expect to be pointed out during review and patched up before merging. I think you could notice that this pull request combines lots of different code from different authors in our team and previously added in different times when they were actually needed. So some unification is definitely to ask but I don't think will be hard to accomplish and this is what reviews are for - to discover maintainers' minimum uniform interface requirements (and leave some divergence that will be unavoidable aside).
> 

I didn't noticed different authors but i do see different levels of automatic transformation of input params (which probably better describes what I was trying to say before. Some functions transform the flags and trying to be safe, some are not). It's hard to decide but in case we can not guarantee safety I would prefer not trying to be too smart as it might give people false impression the utility handles security for them. It does not and the underlying session uses raw bash and should be only used by people with some level of knowledge of the commands they are trying to execute.

> > I can create a repo for you and make you an admin of that, if you are interested, but I don't like the idea of having this in the core aexpect (note ideally I'd extract and publish the `remote.py` separately as well, but it's already there...)
> 
> Why would the `remote` be considered separate from aexpect? The entire point of our previous arguments in the past was that it makes more sense in aexpect than in Avocado VT and I think it extends the `client` module quite nicely adding the file transfer and session login/prompts support. The same can now be said about the `remote_door` module which extends this further with remote function execution and remote object method calls. But letting all of this aside, do you think that a separate project would make things easier? One thing we seem to have learned from the past is that fragmenting projects into too many small ones adds a lot of deployment, versioning, and maintenance overhead resulting in hours of packaging, debugging and testing of version compatibility and so on.
> 

Although I mentioned separate repo, it doesn't have to be that way. Avocado also hosts optional plugins inside the main repo and they can be brought by installing "avocado-framework-plugin-$NAME". Something like that would be nice for aexpect. My reason is that I'm using (and promoting) aexpect as stable and simple utility for automation. Some developers at RH already caught it up and are using it for rebooting machines over serial console, or other simple tasks. I'm also using it in other projects but the cleverer it gets the bigger and potentially more dangerous it gets.

I consider aexpect something like cat. It performs well and allows basic operations. One could use tail for similar tasks and even more, but it is more complex and potentially dangerous. And then you have visual studio, which also can perform file operation, allows all kinds of extra features, but is heavy-weight and aimed for different usecases. Now the question is, is aexpect growing to be closer to tail or to vs?

In my ideal world we'd have probably just 1 repo, but 3 target packages "aexpect", "aexpect-remote" and "aexpect-ops". Would this work for you?

> > Anyway @clebergnu, @lmr please take a look at this and if you decide it belongs here, feel free to give it a try.
> 
> Another point I want to add here is that we find lots of contributors on the Avocado and Avocado VT side reimplementing a lot of these functions on top of a session again and again which confirms the need to have single reusable source that everybody could instead simply spend their time extending and saving the rest of their costs. As these utilities depend only on aexpect and all they require is a session I don't think it makes sense to keep them in Avocado or Avocado VT repos whcih specialize in something else.
> 

And I am honestly sad about that. I have pinged people from avocado-vt to move up the priority of your PR to finally move avocado-vt to aexpect-remote.

> Regarding a separate aexpect repo, can you elaborate more on what are the benefits from keeping current aexpect to the minimal code base and functionality it currently has? I would assume reduced maintenance cost and contributor activity is one of them but how would this compare to the cost of managing a separate repo and integrating the two mentioned above? The aexpect repo has not been very active in the last years but this could be viewed both as something good and something not so good (example of the second being contributors that did not want to use it because they thought the project has been abandoned at least in their replies to me).
> 
> So all in all I would be glad to hear your reasoning regarding keeping an already small project like aexpect bare bone and core-only since at least in my opinion it would rather make sense to split into multiple repos any projects that are much larger than this so that the costs from the splitting are outweighed by the benefits of separate maintenance.

I mentioned it earlier, basically size and security. Core aexpect is very stable, proven to work in many applications and it's API is clearly "dangerous". By adding upper layers people might get false impression of it to be higher-level safe tool, which it is not or it would take a long time to ensure and keep testing the security of it's layer.

clebergnu (changes requested):
Hi @pevogam, thanks for taking the time to compile and submit those utilities.

I won't try to get into each of the points discussed so far, but I'll try to optimize the discussion by talking about the most important ones IMO.

First, about whether this module fits well enough in the aexpect repo.  My take is that it does extend aexpect in ways that are useful to some people and use cases.  IMO, as long as the code matches the project's quality standards, including passing review on its technical merits, and does not add maintenance burden, it deserves to go in.  A project's scope shouldn't be so set in stone, but should respond to what its users need from it.

Second, the organization of aexpect's first level "aexpect" directory never seemed like a "core" module to me.  In fact, this is the location from where all API users actually import code from.  I can't find a difference between `from aexpect import remote` and `from aexpect import ops`.

Finally, a new topic related to maintenance: it would be great if tests could be added.  I understand these are simple wrappers, but as shown in the issue found during review, users and maintainers would benefit from having some coverage there.

ldoktor:
> Hi @pevogam, thanks for taking the time to compile and submit those utilities.
> 
> I won't try to get into each of the points discussed so far, but I'll try to optimize the discussion by talking about the most important ones IMO.
> 
> First, about whether this module fits well enough in the aexpect repo. My take is that it does extend aexpect in ways that are useful to some people and use cases. IMO, as long as the code matches the project's quality standards, including passing review on its technical merits, and does not add maintenance burden, it deserves to go in. A project's scope shouldn't be so set in stone, but should respond to what its users need from it.
> 
> Second, the organization of aexpect's first level "aexpect" directory never seemed like a "core" module to me. In fact, this is the location from where all API users actually import code from. I can't find a difference between `from aexpect import remote` and `from aexpect import ops`.
> 

OK, I understand the point that we already have another "uglier" modules in there. I hope I'll find some time to come-up with a simple to use way to separate them (ideally multiple pip packages generated out of a single source-tree) but until then let's get this in. I mean it is useful, it is way cleaner than remote and Plamen (and others) spent already way too much on updating it.

> Finally, a new topic related to maintenance: it would be great if tests could be added. I understand these are simple wrappers, but as shown in the issue found during review, users and maintainers would benefit from having some coverage there.

Yep, selftests are always welcome, although even the core aexpect contains only a bare minimum selftests.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    Invoke ``grep`` on guest searching for *expr* in *path*. Throws a
+    ``TypeError`` in case the API assumptions are violated.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str expr: search expression
+    :param str path: file to search
+    :param bool check: whether to quietly run grep for a boolean check
+    :param str flags: extra flags passed to ``grep`` on the command line
```
ldoktor, line 76: How much deep do you want to go in abstracting here? I'm usually in favour of simplicity, rather than complex logic and high level default usage with unmodified low-level tweaks if needed.

In this case we have high level API `grep(session, expr, path)`. Optionally we can check status, which is still high-level. Good so far. Then we have `flags`, which is something lower-level and it is specified as `string` and is processed by automatic logic to prefix each one with `-`. This works well with single-letter options like `oP` pr `q`, but does not really fit the long arguments (`--line-regexp`). One could try using lists, which would indeed iterate through strings, but it would silently ignore the option as `-` is not alpha-numeric.

Personally I'd prefer no modifications at all, requiring the users to pass the lower-level arguments themselves, or full-blown API that would take care of translating options (probably list of well defined supported options). This seems like a half-way there and it limits the usage.

Note: I used a long text description here as I see similar issues elsewhere and I'd like to use this first occurrence to properly explain my view and negotiate some agreement on a style for the whole module.

ldoktor: Note that one option would be to specify the high-level API by keywords (like `expr`, `path`, `check`) and add `extra_params` to each command, that would be passed directly to the command with no chacks at all, allowing advanced tweaking of that command and would be common in all commands (that consist of a single call).

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :type session: ShellSession
+    :param str filename: full path of file that should be hashed
+    :param str method: method for hashing
+    :returns: hash of file, a hex number
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if hash command failed (e.g. file not
+             found) or resulting output does not have expected format
+    :raises: :py:class:`ValueError` if given method is not (yet) supported
+    """
+    if method not in ['md5']:
```
ldoktor, line 259: I'm tempted to suggest using `{method}sum` command and returning the error in case the command does not exists. The naming is quite standard and while this module is not really safe and allows any kind of injections, we might not need to care about yet another one. What do you think?

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    cmd = f'tar -C {quote(target_dir)} --strip-components=1 -xapf {quote(tarball)}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to extract {tarball} to {target_dir}: {output}')
+
+
+def ls(session, dir_name):  # pylint: disable=C0103
```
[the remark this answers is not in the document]
ldoktor: This is close to what I wrote at the beginning. IMO the basic API should be simple to use and cover the most common usage, while it would be nice to allow people to tweak the way they want. Here I'd expect: `ls(session, target=None, args='-1UNq')`. This way people could use `ls(session)` as well as completely change the view by `ls(session, '/foo', '--color')` and skip the default arguments. (note I do not say it's the best solution or that you have to adopt it everywhere. But for the raw commands it'd make sense to me)

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :returns: file names matching pattern, including path (like glob.glob)
+    :rtype: [str]
+    :raises: :py:class:`RuntimeError` if tar ls-command returned non-null and
+             non-2 status
+
+    Much like :py:func:`glob.glob` but implemented using `ls`.
+
+    The glob matching is done locally, not with `ls` since quoting also makes
+    the glob characters '*' and '?' regular characters, so are not interpreted
+    any more in shell.
```
ldoktor, line 344: This is not really explicit but it won't won't allow globs like `/*/foo` to get all `foo` files from any root directory. So, provided the implementation stays, it'd be better to explicitly expect `glob(session, path, file_glob)` to avoid surprises.

I think it'd be better to rather implement `raw_ls(session, raw_pattern, args='-1UNq')` which would execute the ls without quotation, therefore allowing the server-side globs.

If you still think the client-side filtering is needed, consider the use of `fnmatch`. I think it would be more standard way (although then we could simply drop the whole function as it'd result in `return [_ for _ in ls(sessionos.path.join(path) if fnmatch(_, file_glob)]`)

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    cmd = 'cat ' + quote(filename)
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to cat {filename}: {output}')
+    return output
+
+
+def tempdir(session):
```
ldoktor, line 384: Did you consider allowing the `template` argument?

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+extracting *atime*, *ctime*, and *mtime*, etc exist.
+
+test:
+Allows for testing existence, permissions and other properties of files.
+
+file ops:
+Simple file operations executed through sessions. Most functions boil down to
+executing simple commands like ls, tar, md5sum, etc. If these operations return
+non-0, a :py:class:`RuntimeError` is raised, containing the commands's error
+message. All functions :py:func:`shlex.quote` their args for better security.
```
ldoktor, line 46: It'd be nice to explain here that this module runs commands on your raw ShellSessions and although we do usually (or always?) quote the main arguments, we do not perform complex checks for escapes and that using of this module might result in data loss or worse.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to stat
+    :param bool human_readable: whether to use human readable format or seconds
+                                since Epoch
+    :returns: standard output of stat(1) querying the size of ``path``
+    :rtype: int
+    """
+    return int(stat(session, path, fmt="%s" if human_readable else "%S"))
```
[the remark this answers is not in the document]
pevogam: Nice catch! I guess the relevant flag symmetries were stretched too far here.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to test
+    :param str flag: Flag passed to test
+    :returns: status of test(1) on ``path``
+    :rtype: int
+    """
+    if not isinstance(flag, str) or flag[0] != "-":
+        raise RuntimeError(f"{flag} is not a valid flag string for test(1)")
+
+    return int(session.cmd_status(f"test {flag} '{path}'"))
```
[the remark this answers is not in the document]
pevogam: Done, all in and fully sold for it being more pythonic.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+def is_directory(session, path):
+    """
+    Check if a directory exists on a given guest.
```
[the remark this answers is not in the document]
pevogam: Indeed, this is a remnant from the code base where these ops were used only for virtual machines. Corrected.

aexpect/ops.py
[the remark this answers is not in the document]
pevogam: Definitely for the mechanism approach as well as adding as much flexibility of usage at an earlier stage as possible. For some more context that you both probably figured out by now: Many of these functions originate at different times (even different years) and implement only as much as was actually needed at the time of writing them in order to keep costs from extra development low (imagining these with tens of other tasks to cross of at the same time).

Even now contributing and merging these would result in a net positive effect for reuse by others and then based on incrementalism the ones that are of greater importance will probably see further extensions and improved applicability. This being said of course I don't mind generalizing some here and now especially if this only requires some minutes development time and just a few small extra steps. Will also make sure to retest the complete module with all our available tests once it stabilizes from a review point of view.

aexpect/ops.py
pevogam: Will do, I hope if you spot any important enough case when we don't `shlex.quote` still and will be good if we do despite this disclaimer, you will let me know.

aexpect/ops.py
pevogam: I agree, I will use `flags` here and unify the approach with a few other locations. I could then add a more universal `flags` argument once I refactor some of the internal command handling in a follow up pull request.

aexpect/ops.py
pevogam: I think it sounds good considering the scope, approach, and current level of implementation of the functions.

aexpect/ops.py
pevogam: Sounds like a quick and easy to add extension, done.

pevogam: Will also do so for `tempfile`.

aexpect/ops.py
pevogam: Inspected the function in detail now and frankly, it seems completely redundant. Relying on the server side glob is also more reliable and reproducible when it comes to debugging so I will simply add `is_glob` argument to the directory listing function.

pevogam:
Hi @clebergnu and @ldoktor, thank you both for spending some of your valuable time on this. I have updated the pull request with all of your comments. I realize and fully agree the code is not of the best quality and I apologize if I am wasting your precious brain cycles on this.

My approach and proposal on how to proceed here is rather simple: we try to satisfy your requirements for some minimum quality in order for the code to become available for others and then we can have any or lots of further improvements to make it more proper and satisfactory depending on demand, i.e. we can always redirect devs that would otherwise reinvent the wheel to solve their own needs as contributors here. In this way it can also be of less effort for you and we can have the highest rates of availability and opportunity for further development with the minimum needed amount of effort.

On my side I will make sure to add some follow up contributions regarding this module as potential subpackage that would address larger refactoring requests and am always available to help with maintenance if you so desire.

aexpect/ops.py
ldoktor: Thanks, I don't think we should. With advanced cases people should take the responsibility themselves...

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    cmd = 'cat ' + quote(filename)
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to cat {filename}: {output}')
+    return output
+
+
+def make_tempdir(session, template="tmp.XXXXXXXXXX"):
```
ldoktor, line 340: I'd suggest using `template=None` here as the default for `mktemp` is to use `TMPDIR` while here it would be inside the current direcotry.

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    and target can be and what behavior to expect.
+    """
+    cmd = f'mv {quote(source)} {quote(target)}'
+    status, output = session.cmd_status_output(cmd)
+    if status != 0:
+        raise RuntimeError(f"Failed to move {source} to {target}: "
+                           f"{output.strip() if output else ''}")
+
+
+def copy(session, source, target):
```
ldoktor, line 404: Here it'd really make sense to allow `flags` (eg. for `-r`)

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    cmd = f'mktemp {template}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to create temporary file: {output}')
+    return output
+
+
+def move(session, source, target):
```
ldoktor, line 384: The same here (as in copy, the `flags` argument) especially for `-f`.

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    and target can be and what behavior to expect.
+    """
+    cmd = f'cp {quote(source)} {quote(target)}'
+    status, output = session.cmd_status_output(cmd)
+    if status != 0:
+        raise RuntimeError(f"Failed to copy {source} to {target}: "
+                           f"{output.strip() if output else ''}")
+
+
+def remove(session, path):
```
ldoktor, line 424: And here the same (`flags` argument)

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    if output:
+        output = output.strip()
+    if status == 2:      # probably just nothing found
+        return []
+    if status != 0:
+        raise RuntimeError(f'Failed to ls {path}: {output}')
+    return output.splitlines()
+
+
+def cat(session, filename):
```
ldoktor, line 317: The `flags` could be useful here as well (eg. for `-v`) but it's not must-have in this case.

aexpect/ops.py
ldoktor: Sure, can be added later although try not to break the existing API.

aexpect/ops.py
pevogam: Hmmm I made sure that the default is exactly the one displayed as the default of `mktemp` when not explicitly provided but using `None` will at the very least make sure updates of the default there propagate and delegation of as much as possible is in the end the better choice so I will go this way.

aexpect/ops.py
pevogam: Alright, let's cover at least the few most obvious choices here

aexpect/ops.py
pevogam: Nice, I added it here too before reading the rest of your comments.

aexpect/ops.py
pevogam: Alright but better to just add to the most critical cases right now since later refactoring could easily make flags available for every command and this might duplicate the effort.

aexpect/ops.py
```diff
@@ -0,0 +1,439 @@
...
+    """
+    cmd = f'tar -C {quote(target_dir)} --strip-components=1 -xapf {quote(tarball)}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status != 0:
+        raise RuntimeError(f'Failed to extract {tarball} to {target_dir}: {output}')
+
+
+def ls(session, path, is_glob=False, flags="-1UNq"):  # pylint: disable=C0103
```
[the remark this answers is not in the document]
pevogam: I was wondering about a proper name here too and thought about using some of names indicating quotation but my problem with this was that it doesn't indicate immediately the rather widely spread glob expansion usage. OTOH a user might expect path expansion by default to begin but then this would contradict with the quote usage for safety and space containing paths handling. So let's stick to `quote_path` and generalize it to one or two other file operations but stay at that for the time being.

About the args order I would definitely prefer it the `flags` that will likely end up being shared by all functions's signature will be the last one with more function-specific arguments coming beforehand.

pevogam:
> > 
> 
> As for me the selftests would be nice but not mandatory at this point. Mocking could consist of simply passing a mock object and checking the output values (ensuring quotation is enforced, arguments are transformed, expected calls are being performed and data are being processed correctly)

This is what I would expect via mocking too but

1) I have had similar tests for a similarly "shallow" code before and the test code is consists almost of a single line that tests the one-line implementation and thus almost black-box-reimplements it - something that I am not particularly fond of. There is the coverage based on actually "running" the code that is needed for interpreted languages like python here but that is then much better covered by a functional test where any changing API of or resulting incompatibility with the underlying binaries will be easily detected (one of the main points of usefulness here) and anything else would require one to update the mock object's validation all the time in order to be up to date with the changing binaries. Granted, this is not exactly the case here consider the age of the cli interface of these binaries but I guess my main point is that the unit tests have to be more carefully considered in order to actually be useful and not fragile, having to be updated every few commits.

2) The signature and output of functions is still not standardized well enough in order to validate sufficient consistency with a pass grade. This is better looked at after contributing more functions and reaching a more final shape of the session ops module.

pevogam:
I tried pushing on most recent master branch to trigger the CI again (didn't trigger within 6 hour timeout upon previous push) but it seems it still won't start. I will however run our functional tests or a good portion of them with the current changes just in case.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    Wrapper for ``test``.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str path: path to the filesystem entry to test
+    :param str flags: flags passed to test
+    :returns: status of test(1) on ``path`` as True or False
+    :rtype: bool
+    """
+    return session.cmd_status(f"test {flags} '{path}'") == 0
```
ldoktor, line 164: The `path` should use quotation.

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :param str target: full path of new file or directory
+    :param bool quote_path: whether to quote (escape) or not (expand) the path
+    :param str flags: raw flags to pass to cp command
+    :raises: :py:class:`RuntimeError` if cp returns non-zero exit status
+
+    Calls `cp source target` through a session. See `man cp` for what source
+    and target can be and what behavior to expect.
+    """
+    cmd = f'cp {flags} {quote(source)} {quote(target)}' if quote_path \
+        else f'cp {flags} {source} {target}'
```
ldoktor, line 276: Wouldn't it be easier to follow with pure old if? (I'm not a fun of backslashed long expressions)

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    LOG.info("Replacing {pattern} with {replacement} in {filename}")
+    # for lines matching pattern, replace and print them (the p command won't work)
+    cmd = f"""sed -ri '/{pattern}/{{
+s//{replacement}/g
+w /dev/stdout
+}}' {filename}"""
+    output = session.cmd(cmd)
+    # make sure we have at least one line replaced
+    assert len(output.splitlines()) > 0, f"sed did not find any matches for pattern {pattern}"
```
ldoktor, line 396: It's safer to raise exceptions as asserts are stripped when compiling with optimizations (although we do use it on a few places in aexpect already so take this as a mild recommendation)

ldoktor: Also thank you for an interesting insight to verifying sed (although I never needed such feature it seems useful)

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :rtype: [str]
+
+    Just like :py:func:`os.listdir`, does not include file names starting with
+    dot (`'.'`)
+    """
+    cmd = f'ls {flags} {quote(path)}' if quote_path else f'ls {flags} {path}'
+    status, output = session.cmd_status_output(cmd)
+    if output:
+        output = output.strip()
+    if status == 2:      # probably just nothing found
```
ldoktor, line 230: Ls reports `2` when any of the expressions fails (eg. due to permission reasons). I'm wondering whether silent return of empty list is the right solution. For example "ls /tmp/*" reports many files and several "permission denied" messages and returns 2, but this code reports empty list with no warning, no failure...

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    :type session: ShellSession
+    :param str filename: full path of file that should be hashed
+    :param str method: method for hashing
+    :returns: hash of file, a hex number
+    :rtype: str
+    :raises: :py:class:`RuntimeError` if hash command failed (e.g. file not
+             found) or resulting output does not have expected format
+    """
+    cmd = f"{method}sum"
+    expect_len = 32
```
ldoktor, line 442: The expect_len should be only checked for `md5` as other cyphers are using different lengths (or we can build a map, but I don't think it'd be that necessary)

aexpect/ops.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    Extract tarball to given dir from a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str tarball: full path of tar file that should be extracted
+    :param str target_dir: name of directory where tarball should be extracted
+    :raises: :py:class:`RuntimeError` if tar command returned non-null
+    """
+    cmd = f'tar -C {quote(target_dir)} --strip-components=1 -xapf {quote(tarball)}'
```
ldoktor, line 474: The use of `--strip-components` can be tricky as on relative path it skips all files from the current directory (eg. `tar -jcf archive.tar.bz2 myfile` contains myfile but it won't get extracted by this command...

aexpect/ops.py
pevogam: IIUC you mean to use `shlex.quote` here, I have also added it to a few more places.

aexpect/ops.py
pevogam: Yes, I had the same concern and found some of our tests would rely on this status to conclude the path is empty. Will consult with the author and if they confirm I would also prefer to simply drop this strange handling and error out as everywhere else.

aexpect/ops.py
pevogam: I also don't like backslashed expressions but having one liner if statements fits better when defining a single value so at least on my side I would prefer to keep it as is. Tell me if you insist on this change, the current statement seems readable and shorter than full blown if clauses to me and might even be refactored through some standardization on the quote behaviors in commands later on.

aexpect/ops.py
pevogam: Removing assertions for the purpose of optimization is something I typically strongly advise against, adding self-validation within the executing code is something very much needed to enhance the language's self-validation in these cases and and I doubt assertions like these are critical performance drag.

Regarding the self check of sed, this is another use we could have from such commands - use them to have additional sanity checks in place and post-processing that fits the aexpect specific usage.

aexpect/ops.py
pevogam: Indeed, sounds a bit of an overkill and the length check is not that useful for the purpose of the function which is to generate a hash for a given file. Validating the hash characters should be enough in this case and still general enough.

aexpect/ops.py
pevogam: Fair enough, will use `flags` argument here.

aexpect/ops.py
ldoktor: It can stay as is although I disagree with the readability, but it's just a matter of taste :-)

aexpect/ops.py
ldoktor: Yes please, especially the fact that it ignores the output and silently returns empty list as if nothing happened smells like troubles to me.

aexpect/ops.py
ldoktor: "Assertions are programmer aids, not error handlers for your business logic" - https://www.inspiredpython.com/article/assertions-and-exceptions-are-not-the-same

aexpect/ops.py
pevogam: Done, thanks for adding more consistency observations

aexpect/ops.py
pevogam: Unless one could calculate the entropy of the two statements and prove your version contains more information and/or correlates the thought invocations when parsing the line with the typical neural pathways of a human reader ^^

aexpect/ops.py
pevogam: I never intended to say they are the same but the current code does state otherwise and I moved this to similar RuntimeError handling logic with other functions.

aexpect/ops.py
pevogam: My still ongoing problem with this is how `ls` would treat glob expansion when there are no items in the expected list:
```
[root@jupiter intra2net-qa-deps]# ls /tmp/no-files*
ls: cannot access '/tmp/no-files*': No such file or directory
[root@jupiter intra2net-qa-deps]# echo $?
2
```
It would be much better if `ls` would not mix this status with as a "serious trouble":
```
   Exit status:
       0      if OK,

       1      if minor problems (e.g., cannot access subdirectory),

       2      if serious trouble (e.g., cannot access command-line argument).
```
In other words what one would expect when expanding a list of paths that is zero is to get an empty list back. Do you perhaps have any other suggestion on how to go around this?

pevogam: I think the only option we have here is to manually parse the stderr for "No such file or directory" even though it seems quite crude for a solution. Do you have any better idea?

pevogam: I might have found an option in bash like
```
$ shopt -s nullglob
$ echo "$(doc/xxxx*/abc/*)"
$ echo "${list[@]}" 

list=(doc/*)
echo "${list[@]}"
doc/api doc/api_create.py doc/api_entry.html doc/api_serve.sh doc/guides doc/README.md doc/references
```
but it depends on the shell type and is different for zsh, ksh93, etc.

pevogam: Not sure again but how about simply restoring the glob function in the module and not relying on `ls` there but on something simple like
```
$ echo doc/*
doc/api doc/api_create.py doc/api_entry.html doc/api_serve.sh doc/guides doc/README.md doc/references
$ echo doc/xxxx*/abc/*

```
?

aexpect/ops.py
```diff
@@ -0,0 +1,475 @@
...
+executing simple commands like ls, tar, md5sum, etc. If these operations return
+non-0, a :py:class:`RuntimeError` is raised, containing the commands's error
+message. All functions :py:func:`shlex.quote` their args for better security.
+
+utilities:
+More complex linux utilities for tarball extraction, file hashing, etc.
+
+..warning:: This module runs commands on raw shell sessions and although we usually
+    quote the main arguments, we do not perform complex checks for escapes. It is
+    meant only for some conveniene in simple enough use cases and using it might
```
lmr, line 49: Nit: conveniene/convenience

lmr (commented):
Hi @pevogam, thank you so much for the library.

I read the discussion so far, and here are my 2 cents about your code:

1. I generally like the idea of having these shell functions (or operations) as a library, as they are widely used in a variety of situations. Having them neatly packaged is a good idea.
As @clebergnu mentioned, as long as we meet the necessary criteria for inclusion (as far as I can see the code generally is clean and documented), it makes for a good inclusion.

2. Creating an extra separate library to interact with aexpect might be too much granularity, as this is fairly small and will be mostly concentrated in the client library. We could think of an `aexpect-extensions` library if we ever get too many extra functions, but that doesn't seem very likely at the moment.

3. As far as code `organization` style goes, all these functions take session as the first parameter. Have you considered making the functions an object that inherits from `session`? That's something like:

```
import aexpect

session = aexpect.ShellOpsSession("sh")
session.stat(path="/tmp/a")
session.copy(src="/tmp/a", dst="/tmp/b")
```
That saves a bit of typing and it seems like a reasonable layering of functionality.

Of course if @clebergnu and @ldoktor disagree, please let me know if I'm off here.

Once again, thanks for your contribution!

aexpect/ops.py
pevogam: Good catch! Fixed.

pevogam:
Hi @lmr,

>     1. I generally like the idea of having these shell functions (or operations) as a library, as they are widely used in a variety of situations. Having them neatly packaged is a good idea.
>        As @clebergnu mentioned, as long as we meet the necessary criteria for inclusion (as far as I can see the code generally is clean and documented), it makes for a good inclusion.

Thanks both for chiming in with views on this, I agree that the current functionality is quite skinny when it comes to wrapping logic around the shell command strings but it could easily grow into some extended and extra useful logic too.

>     2. Creating an extra separate library to interact with aexpect might be too much granularity, as this is fairly small and will be mostly concentrated in the client library. We could think of an `aexpect-extensions` library if we ever get too many extra functions, but that doesn't seem very likely at the moment.

Yes, I am all in for separating if this becomes big enough later on but just like you I think that at the moment creating separate project infrastructure is a bit of an overkill.

>     3. As far as code organization style goes, all these functions take session as the first parameter. Have you considered making the functions an object that inherits from `session`? That's something like:
> 
> 
> ```
> from aexpect.remote.ops import ShellOpsSession
> 
> session = ShellOpsSession("sh")
> session.stat(path="/tmp/a")
> session.copy(src="/tmp/a", dst="/tmp/b")
> ```
> 
> That saves a bit of typing and it seems like a reasonable layering of functionality.
> 

I can see two problems with this:
1) Typing `session.copy(A, B)` is not much shorter than typing `ops.copy(session, A, B)` (about 3+ chars difference).
2) More importantly, adding more functions later on would require us to move from `ops.py` module to `ops` subpackage with the current sections becoming separate modules as was recommended by some of our previous reviewers. Extending a class over multiple module seems somewhat incorrect and it would make more sense for people to just import an ops subpackage or modules and functions thereof. Note that this is still an intermediate stage that would take a while before coming to the need to talk about a separate project of course.

ldoktor (approved):
I'm still not happy about this not being a separate module, but it looks good and others are fine having it in-tree, so I think it's ready for inclusion. Maybe one thing to consider is the `strip` in the `cat` method. I do understand why it's there, but maybe it'd be safer to leave it on the user. Anyway let's not block this anymore and include this first implementation, we can tweak it later, if necessary.

ldoktor:
Well it re-runs with the same base commit, anyway passes local "make check" so let me merge this.

pevogam:
Hi @ldoktor and thanks all for all the reviews here,

> Well it re-runs with the same base commit, anyway passes local "make check" so let me merge this.

Hmm but as I wrote above the situation with the `ls` and status 2 is still not resolved. Should I write an issue about it or propose a pull request? Could you still comment on https://github.com/avocado-framework/aexpect/pull/98#discussion_r1003257178 above?

aexpect/ops.py
ldoktor: Hello @pevogam, I'm sorry for not noticing this pending request earlier. Now the function is called `ls` so I'd expect it to act as ls, ideally without corner cases, if possible. But you are the author and most certainly the biggest user, so I think it makes sense for you to define the rules here. I'm fine with anything systematic, ideally without exceptions (or just minor corner cases).

pevogam: I will look into this and do some testing in advance then possibly push a new pull request. Just FYI, this is the last open issue regarding the current merged that I am aware of and instead of opening a new issue I will simply push the corresponding pull request next, I have no other pending issues to register at aexpect on my side at least.

pevogam: Follow up on https://github.com/avocado-framework/aexpect/pull/113

PR #99, opened 2022-09-24
Simpler ns daemon and current linter fixes

pevogam:
The previous approach of using a variable initialied only on given
condition after verifying the same condition seems to be all too
convoluted for linters so let's simplify it until they are happy.

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -613,17 +613,16 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
         pyrod_running = True
 
     # name server
+    ns_daemon = None
     try:
         ns_server = Pyro4.locateNS(host=host, port=port)
         LOG.debug("Pyro4 name server already started")
-        nsd_running = True
     # network unreachable and failed to locate the nameserver error
     except (OSError, Pyro4.errors.NamingError):
         from Pyro4 import naming
         ns_uri, ns_daemon, _bc_server = naming.startNS(host=host, port=port)
         ns_server = Pyro4.Proxy(ns_uri)
         LOG.debug("Pyro4 name server started successfully with URI %s", ns_uri)
-        nsd_running = False
 
     # main retrieval of the local object
     module = importlib.import_module(object_name)
@@ -674,7 +673,8 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
     if not pyrod_running:
         loop = DaemonLoop(pyro_daemon)
         loop.start()
-    if not nsd_running:
+    # if name service daemon is running
+    if ns_daemon is not None:
         loop = DaemonLoop(ns_daemon)
         loop.start()
     # we should register to the name server after entering its loop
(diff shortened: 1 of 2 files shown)


pevogam:
This fixes the issue with the naming but there is some global check failing in one variant that I am not sure where to find (it would be nice if the CI checks have at least a single docstring or sentence of what is being checked and hot to fix it, currently it only says "global check".

aexpect/remote_door.py
```diff
@@ -616,6 +616,7 @@ def share_local_object(object_name, whitelist=None, host="localhost", port=9090)
     try:
         ns_server = Pyro4.locateNS(host=host, port=port)
         LOG.debug("Pyro4 name server already started")
+        ns_daemon = None
```
ldoktor, line 619: It would be better to initialize the `ns_daemon` outside the `try/except` block. I know the except block currently results into `nsd_running == False` so the `ns_daemon` would not be checked, but if the code changes somehow everything could break.

ldoktor:
Hello Plamen, thank you for taking care of the CI issues. If I understand the second warning correctly, they removed the W1618 from the list of warnings. So something like:

```diff
diff --git a/Makefile b/Makefile
index 0158a12..867ee37 100644
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ rpm-release: srpm-release
        mock -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "rel_build 1" --rebuild BUILD/SRPM/python-aexpect-$(VERSION)-*.src.rpm
 
 check: clean
-       inspekt checkall --disable-lint W1618,R0205
+       inspekt checkall --disable-lint R0205
        $(PYTHON) setup.py develop $(PYTHON_DEVELOP_ARGS)
        $(PYTHON) setup.py test
 
```

Should do and even older (F36) pylint is not complaining about W1618s in our code...

ldoktor:
Oh, could you please push this without the session ops commit so we can discuss that one in the #98?

aexpect/remote_door.py
pevogam: Sounds good, just wanted to minimize the changes and still make the linters happy.

aexpect/remote_door.py
```diff
@@ -674,7 +673,8 @@ class ModuleObject:  # pylint: disable=R0903
     if not pyrod_running:
         loop = DaemonLoop(pyro_daemon)
         loop.start()
-    if not nsd_running:
+    # if name service daemon is running
+    if ns_daemon is not None:
```
ldoktor, line 677: Well that is actually even better :-) Thank you.

pevogam:
Hi @ldoktor,

> Should do and even older (F36) pylint is not complaining about W1618s in our code...

I integrated this in a single commit with the "fix linter issues" topic, see the commit message for details. Linters pass now!

> Oh, could you please push this without the session ops commit so we can discuss that one in the https://github.com/avocado-framework/aexpect/pull/98?

Indeed, a small slip up because I combine all recent pull requests in my testing and some of them are indispensable in our everyday use of aexpect, sorry about that - fixed now.

ldoktor (approved):
Thank you, works well

PR #101, opened 2022-09-28
Drop unexpected and unwanted re.sub string postprocessing

pevogam:
Using re.sub would result in some special characters like backslashes being processed despite us providing repr() of the metaprogramming embedded values. This is only a problem in some rare cases but as we want any code to be included in a remote control file as-is in its raw form, repr() is best utilized with a more standard string replace() call.

To further illustrate what is meant here:

>>> print("b = " + repr(p["tmp"]))
b = 'C:\\'
>>> print(re.sub(r"a = .*", "b = " + repr(p["tmp"]), "a = abc"))
b = 'C:\'

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -90,6 +90,7 @@ LOG = logging.getLogger(__name__)
 #   REMOTE UTILITIES / GENERATED CONTROLS   #
 #############################################
 
+
 #: default location for template control files
 SRC_CONTROL_DIR = "."
 #: default location for dumped control files
@@ -208,6 +209,7 @@ def run_remotely(function):
         return re.search("RESULT = (.*)", full_output).group(1)
 
     return wrapper
+
 
 ##################################
 #   STATIC (TEMPLATE) CONTROLS   #
@@ -341,9 +343,10 @@ def set_subcontrol_parameter(subcontrol, parameter, value):
     .. warning:: The `subcontrol` parameter is control path externally but
         control content internally after decoration.
     """
-    return re.sub(f"{parameter.upper()}[ \t\v]*=[ \t\v]*.*",
-                  f"{parameter.upper()} = {value!r}",
-                  subcontrol, count=1)
+    line = re.search(fr"{parameter.upper()}[ \t\v]*=[ \t\v]*.*", subcontrol)
+    line = "" if line is None else line.group()
+    # re.sub does undesirable post-processing of the replaced string
+    return subcontrol.replace(line, f"{parameter.upper()} = {value!r}")
 
 
 @set_subcontrol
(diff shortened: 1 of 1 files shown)


pevogam:
Hmmm, this time the CI was simply canceled after 6 hours of not starting at all.

aexpect/remote_door.py
```diff
@@ -341,9 +343,10 @@ def set_subcontrol_parameter(subcontrol, parameter, value):
     .. warning:: The `subcontrol` parameter is control path externally but
         control content internally after decoration.
     """
-    return re.sub(f"{parameter.upper()}[ \t\v]*=[ \t\v]*.*",
-                  f"{parameter.upper()} = {value!r}",
-                  subcontrol, count=1)
+    line = re.search(fr"{parameter.upper()}[ \t\v]*=[ \t\v]*.*", subcontrol)
+    line = "" if line is None else line.group()
+    # re.sub does undesirable post-processing of the replaced string
+    return subcontrol.replace(line, f"{parameter.upper()} = {value!r}")
```
ldoktor, line 349: Hello Plamen, could you please add a selftest or at least provide the the input parameters that are failing? I tested it by:

```python
remote_door.set_subcontrol_parameter("ASD='D:\\'\nDSA = 'E:\\'", 'ASD', 'C:\\')
```

(removing the `@set_subcontrol` decorator) and the original code reported `"ASD = 'C:\\'\nDSA = 'E:\\'"` while the new version `"ASD = 'C:\\\\'\nDSA = 'E:\\'"`.

I also tried with the `@set_subcontrol` decorator and the original version worked well while this one doubled the back slashes.

pevogam: > Hello Plamen, could you please add a selftest or at least provide the the input parameters that are failing?

Have you tried the small reproducing snippet I provided in the original pull request description? Do you understand from this description what the problem I am trying to fix is? In particular:

> (removing the `@set_subcontrol` decorator) and the original code reported `"ASD = 'C:\\'\nDSA = 'E:\\'"` while the new version `"ASD = 'C:\\\\'\nDSA = 'E:\\'"`.

How did you test the "reported output" here? Was it via `print()` (analogous to the `@set_subcontrol` decorator's writing the string to a file) or perhaps just an interpreter's repr output like
```
>>> reported_string = <some-manupulaton>
>>> reported_string
```
If the string that these functions output is written to a file or printed you will see:
```
>>> import re
>>> re.sub("ASD = 'C:\\'\nDSA = 'E:\\'", "ASD = 'C:\\'\nDSA = 'E:\\'", 'C:\\')
'C:\\'
>>> print(re.sub("ASD = 'C:\\'\nDSA = 'E:\\'", "ASD = 'C:\\'\nDSA = 'E:\\'", 'C:\\'))
C:\
```
The second output when written into a file from a metaprogramming point of view will result in a syntax error.

ldoktor: I used interactive interpreter but print/write works differently, my fault. So yes, this works definitely better and thank you for the confirmation.

pevogam: No worries

ldoktor (approved):
OK, seems to work well, thank you.

PR #102, opened 2022-10-19
Remove six dependency

### LINKED ISSUE #83: Remove six dependency
aexpect is not Python 2 compatible anymore and uses `six` only for a single function (`reraise()`).  Let's remove this dependency to simplify deployment.

danigm:
Fix #83


ldoktor (approved):
Thank you, @danigm, this is the last `six` reference.

PR #103, opened 2022-10-24
Fix remote_door set_subcontrol_* handling of no matches

ldoktor:
In case the parameter is not found we need to return immediately, otherwise we'd be replacing everything with our parameter.

This was introduced in https://github.com/avocado-framework/aexpect/pull/101 and was the source of the selftest hangs.

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -343,10 +343,12 @@ def set_subcontrol_parameter(subcontrol, parameter, value):
     .. warning:: The `subcontrol` parameter is control path externally but
         control content internally after decoration.
     """
-    line = re.search(fr"{parameter.upper()}[ \t\v]*=[ \t\v]*.*", subcontrol)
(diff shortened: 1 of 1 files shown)


ldoktor:
@pevogam could you please verify my fix is correct? IIUC the code should perform no modifications in case the parameter was not set in the original subcontrol file, right? In that case we should not need to run any replacement and simply return.

pevogam:
@ldoktor Nice catch! Is this the reason why the CI was hanging? I thought it was a problem on the GH actions side and never looked deeper into it. I guess the lesson to be learned is that we should always check the CI status even if it is potential noise from external problems. Will include this into some quick end-to-end testing on our side and ACK the pull request as soon as it completes. Thanks!

pevogam (approved):
All works as expected, LGTM

ldoktor:
Thank you for the review. Let me merge this and re-run the other PR checks.

PR #106, opened 2022-11-03
Add a section for system session ops

pevogam:
Currently the focus is on system services which is what some of us users needed so far but it will easily be extended further.

### DIFF
--- a/aexpect/ops.py
+++ b/aexpect/ops.py
@@ -26,7 +26,8 @@ environment (i. e. Cygwin) is present.
 
 This module aims to provide convenient wrappers for butter-and-bread
 functionality on a Linux system. Currently, the implemented functions fall
-under a few categories: ``stat``, ``test``, common file ops, and utilities.
+under a few categories: ``stat``, ``test``, common file ops, common utilities,
+and system utilities.
 
 stat:
 Allows passing an optional format argument (``-c``). Custom wrappers for
@@ -44,7 +45,10 @@ message. All functions :py:func:`shlex.quote` their args for better security.
 utilities:
 More complex linux utilities for tarball extraction, file hashing, etc.
 
-..warning:: This module runs commands on raw shell sessions, and although we usually
+system:
+System related operations like restarting or stopping services.
+
+..warning:: This module runs commands on raw shell sessions and although we usually
     quote the main arguments, we do not perform complex checks for escapes. It is
     meant only for some convenience in simple enough use cases and using it might
     result in data loss or worse.
@@ -473,3 +477,94 @@ def extract_tarball(session, tarball, target_dir, flags="-ap"):
         output = output.strip()
     if status != 0:
         raise RuntimeError(f'Failed to extract {tarball} to {target_dir}: {output}')
+
+
+###############################################################################
+# system
+###############################################################################
+
+
+def start_services(session, services, timeout=60):
+    """
+    Try starting all services in given list.
+
+    :param session: virtual machine session
+    :type session: :py:class:`aexpect.ShellSession`
+    :param services: list of services to modify
+    :type services: [str]
+    :param int timeout: timeout for each service to start
+    :returns: statuses for each modified service
+    :rtype: [int]
+    """
+    statuses = []
+    for service in services:
+        status = session.cmd_status(f'service {service} status')
+        if status == 0:      # is running --> skip
+            LOG.debug(f'Service {service} already started - skipping')
+        elif status == 3:    # is stopped --> ok
+            LOG.debug(f'Service {service} stopped - starting')
+            status = session.cmd_status(f'service {service} start', timeout)
+        else:                # not sure, try restarting
+            LOG.debug(f'Querying status for service {service} resulted in '
+                      f'unexpected return code {status}. Try starting it')
+            status = session.cmd_status(f'service {service} start', timeout)
+        LOG.debug(f'Starting service {service} resulted in status {status}')
+        statuses += [status]
+    return statuses
+
+
+def stop_services(session, services, timeout=60):
+    """
+    Try stopping all services in given list.
+
+    :param session: virtual machine session
+    :type session: :py:class:`aexpect.ShellSession`
+    :param services: list of services to modify
+    :type services: [str]
+    :param int timeout: timeout for each service to stop
+    :returns: statuses for each modified service
+    :rtype: [int]
+    """
+    statuses = []
+    for service in services:
+        status = session.cmd_status(f'service {service} status')
+        if status == 0:      # is running --> stop
+            status = session.cmd_status(f'service {service} stop', timeout)
+        elif status == 3:    # is stopped --> ok
+            LOG.debug(f'Service {service} stopped already')
+        else:                # not sure, try stopping
+            LOG.debug(f'Querying status for service {service} resulted in '
+                      f'unexpected return code {status}. Try stopping it')
+            status = session.cmd_status(f'service {service} stop', timeout)
+        LOG.debug(f'Stopping service {service} resulted in status {status}')
+        statuses += [status]
+    return statuses
+
+
+def restart_services(session, services, timeout=60):
+    """
+    Try restarting all services in given list.
+
+    :param session: virtual machine session
+    :type session: :py:class:`aexpect.ShellSession`
+    :param services: list of services to modify
+    :type services: [str]
+    :param int timeout: timeout for each service to restart
+    :returns: statuses for each modified service
+    :rtype: [int]
+    """
+    statuses = []
+    for service in services:
+        status = session.cmd_status(f'service {service} status')
+        if status == 0:      # is running --> restart
+            status = session.cmd_status(f'service {service} restart', timeout)
+        elif status == 3:    # is stopped --> ok
+            LOG.debug(f'Service {service} stopped - starting')
+            status = session.cmd_status(f'service {service} start', timeout)
+        else:                # not sure, try restarting
+            LOG.debug(f'Querying status for service {service} resulted in '
+                      f'unexpected return code {status}. Try restarting it')
+            status = session.cmd_status(f'service {service} restart', timeout)
+        LOG.debug(f'Restarting service {service} resulted in status {status}')
+        statuses += [status]
+    return statuses


pevogam:
Is the lazy logging formatting really obligatory? I am aware there are optimizations for it but there are also lots of counterargument against it and for using fstrings for logging messages instead. So in short, how hard set in stone is this requirement? I had to convert to lazy formatting for another pull request too but this now makes me explicitly ask about this.

ldoktor:
Hello Plamen, this is exactly what I was afraid of when you introduced the session ops. Yes, these functions are useful, but IMO outside of the scope of aexpect. Very soon you'll have to split the `ops` into multiple submodules to categorize all the command groups, they will grow and the logic will be more complex.

We do already have a nice and aexpect compatible way of handling services in: https://github.com/avocado-framework/avocado/blob/master/avocado/utils/service.py have you considered using that? I know it would bring the whole avocado dependency, but perhaps it'd be better to focus on the separation: https://github.com/avocado-framework/utils I think your `ops` utility would suite well there, what do you say? I think @beraldoleal or @clebergnu could give us an estimate on when this library should be split. (note I'm a big fun of the split and the tools themselves so I might be biased a bit)

pevogam:
Hi Lukas,

> Hello Plamen, this is exactly what I was afraid of when you introduced the session ops. Yes, these functions are useful, but IMO outside of the scope of aexpect. Very soon you'll have to split the `ops` into multiple submodules to categorize all the command groups, they will grow and the logic will be more complex.

I tend to disgree here - pull requests like this one were already mentioned and expected in the original discussion. We never said that the aexpect session ops won't grow any further and this will be a single one-time pull request to begin with so I am not sure why this pull request should be surprising in any way. Just as you said, this should grow into more sections and a potential subpackage, in all of those evolutionary steps I don't see it being large enough to be considered an "external project".

> We do already have a nice and aexpect compatible way of handling services in: https://github.com/avocado-framework/avocado/blob/master/avocado/utils/service.py have you considered using that? I know it would bring the whole avocado dependency, but perhaps it'd be better to focus on the separation: https://github.com/avocado-framework/utils I think your `ops` utility would suite well there, what do you say? I think @beraldoleal or @clebergnu could give us an estimate on when this library should be split. (note I'm a big fun of the split and the tools themselves so I might be biased a bit)

The avocado service utility is based on local command running and is not aware of remote sessions in any way. And while I might be for some command running functional replacement to make it work with aexpect I was intending to push a pull request in the future that introduces session ops for remote sessions to windows which further distinguishes the portfolio of session ops that aexpect can provide.

I aexpect builds a lot of logic that wraps naturally around all of tail/expect/sessions/remote then I don't see why this logic cannot be continued to include useful operations. But then it seems we arrive at the same point where we were before so once again I am up for opinions from others.

ldoktor:
> Hi Lukáš,
> 
> > Hello Plamen, this is exactly what I was afraid of when you introduced the session ops. Yes, these functions are useful, but IMO outside of the scope of aexpect. Very soon you'll have to split the `ops` into multiple submodules to categorize all the command groups, they will grow and the logic will be more complex.
> 
> I tend to disgree here - pull requests like this one were already mentioned and expected in the original discussion. We never said that the aexpect session ops won't grow any further and this will be a single one-time pull request to begin with so I am not sure why this pull request should be surprising in any way. Just as you said, this should grow into more sections and a potential subpackage, in all of those evolutionary steps I don't see it being large enough to be considered an "external project".
> 

I don't deny that, I'm just looking at this slowly becoming `avocado.utils` which should have been extracted a long time ago and I'd like to avoid re-inventing the wheel from a different side.

> > We do already have a nice and aexpect compatible way of handling services in: https://github.com/avocado-framework/avocado/blob/master/avocado/utils/service.py have you considered using that? I know it would bring the whole avocado dependency, but perhaps it'd be better to focus on the separation: https://github.com/avocado-framework/utils I think your `ops` utility would suite well there, what do you say? I think @beraldoleal or @clebergnu could give us an estimate on when this library should be split. (note I'm a big fun of the split and the tools themselves so I might be biased a bit)
> 
> The avocado service utility is based on local command running and is not aware of remote sessions in any way. And while I might be for some command running functional replacement to make it work with aexpect I was intending to push a pull request in the future that introduces session ops for remote sessions to windows which further distinguishes the portfolio of session ops that aexpect can provide.
> 

The `avocado.utils.service` consumes the `run` argument, which can be `process.run` or `session.cmd`.

The extension to windows was actually what I imagined and what lead to the "sigh" at the beginning. The `avocado.utils.service` has a long history. Although it might not be perfect and in the best shape, it contains the logic to deal with init as well as systemd. Adding windows support should not be that hard as implementing it from scratch.

> In general aexpect builds a lot of logic that wraps naturally around all of tail/expect/sessions/remote then I don't see why this logic cannot be continued to include useful operations. But then it seems we arrive at the same point where we were before so once again I am up for opinions from others.

The damage (in my eyes; don't want to start a flame-war) had already be done so I will not attempt to block this. The intention of my answer was to point out these higher-level helpers were already written and I believe there is a better place for similar ones than aexpect. Aexpect is a very neat library and it's purpose should be to perfectly handle basic interaction and the higher layers like systemd, networking, podman and ideally the cmds as well belongs in my opinion elsewhere, for example into the `avocado.utils`.

pevogam:
> > The avocado service utility is based on local command running and is not aware of remote sessions in any way. And while I might be for some command running functional replacement to make it work with aexpect I was intending to push a pull request in the future that introduces session ops for remote sessions to windows which further distinguishes the portfolio of session ops that aexpect can provide.
> 
> The `avocado.utils.service` consumes the `run` argument, which can be `process.run` or `session.cmd`.

This is what I expected you would say - we have lots of similar local utilities taking the same approach. In that case I guess you would imagine a newly created future avocado.utils to have aexpect as direct dependence and perform various operations with both local and remote (aexpect behind the curtain) support. It does seem reasonable of course.

> The extension to windows was actually what I imagined and what lead to the "sigh" at the beginning. The `avocado.utils.service` has a long history. Although it might not be perfect and in the best shape, it contains the logic to deal with init as well as systemd. Adding windows support should not be that hard as implementing it from scratch.

I am not sure if we really imagine the same thing - the windows ops enhance aexpect with the ability to run PowerShell scripts remotely, wraps various other windows commands and capabilities, all of which seems generic enough in my opinion but I guess we already don't exactly agree on the scope of generality of the current session ops to begin with ^^ Just to illustrate: I don't think functionality related strictly to ways of making it possible to run a PowerShell script remotely has much to do with the potential local/remote mix in the potential avocado utils, it seems more like code that enhances aexpect itself and should therefore be applied in the greatest scope of impact (and maximum reusability of other projects). Hence my point that we might imagine different things when it comes to the windows ops still.

> > In general aexpect builds a lot of logic that wraps naturally around all of tail/expect/sessions/remote then I don't see why this logic cannot be continued to include useful operations. But then it seems we arrive at the same point where we were before so once again I am up for opinions from others.
> 
> The damage (in my eyes; don't want to start a flame-war) had already be done so I will not attempt to block this. The intention of my answer was to point out these higher-level helpers were already written and I believe there is a better place for similar ones than aexpect. Aexpect is a very neat library and it's purpose should be to perfectly handle basic interaction and the higher layers like systemd, networking, podman and ideally the cmds as well belongs in my opinion elsewhere, for example into the `avocado.utils`.

I think historically it is still fine to consider an initial growth of utilities in aexpect and then migrate/generalize/integrate those to avocado.utils when the respective project is available. There are lots of plans to collect utilities from various locations to begin with, I don't see why aexpect cannot be one of them. Historically it simply means that aexpect eventually grew its own utilities that were eventually pulled there. In this sense there is no damage that was already done but a normal developments of utilities that happens on a multitude of projects under the avocado umbrella.

aexpect/ops.py
```diff
@@ -473,3 +477,94 @@ def extract_tarball(session, tarball, target_dir, flags="-ap"):
...
+    :rtype: [int]
+    """
+    statuses = []
+    for service in services:
+        status = session.cmd_status(f'service {service} status')
+        if status == 0:      # is running --> restart
+            status = session.cmd_status(f'service {service} restart', timeout)
+        elif status == 3:    # is stopped --> ok
+            LOG.debug(f'Service {service} stopped - starting')
+            status = session.cmd_status(f'service {service} start', timeout)
```
ldoktor, line 563: Is it necessary to distinguish between those? The `service` tool ensures that already.

aexpect/ops.py
```diff
@@ -473,3 +477,94 @@ def extract_tarball(session, tarball, target_dir, flags="-ap"):
...
+        if status == 0:      # is running --> skip
+            LOG.debug(f'Service {service} already started - skipping')
+        elif status == 3:    # is stopped --> ok
+            LOG.debug(f'Service {service} stopped - starting')
+            status = session.cmd_status(f'service {service} start', timeout)
+        else:                # not sure, try restarting
+            LOG.debug(f'Querying status for service {service} resulted in '
+                      f'unexpected return code {status}. Try starting it')
+            status = session.cmd_status(f'service {service} start', timeout)
+        LOG.debug(f'Starting service {service} resulted in status {status}')
```
ldoktor, line 511: This is quite chatty, how about combining those messages into one?

    Service {service} was running
    Service {service} was stopped, start returned {status}
    Service {service} status command returned {stat_status}, start returned {status}

And similarly elsewhere?

ldoktor:
@pevogam as I mentioned earlier, I don't mean to block this, I was just curious about the direction and scope. It seems to be quite ambitious and I think it'd make sense to get in touch with @clebergnu and @beraldoleal to avoid reinventing the wheel as they are also interested in extracting and generalizing the `avocado.utils`. In the mean-time as long as it's just a single file, let's keep this in `aexpect`, but the moment it becomes bigger let's remove it and make a separate `aexpect.ops` sub-project (could be in the same tree, no problem with that)

aexpect/ops.py
```diff
@@ -473,3 +477,94 @@ def extract_tarball(session, tarball, target_dir, flags="-ap"):
...
+    :type services: [str]
+    :param int timeout: timeout for each service to start
+    :returns: statuses for each modified service
+    :rtype: [int]
+    """
+    statuses = []
+    for service in services:
+        status = session.cmd_status(f'service {service} status')
+        if status == 0:      # is running --> skip
+            LOG.debug(f'Service {service} already started - skipping')
```
ldoktor, line 503: logs should use lazy prints rather than f-strings.

pevogam: See https://github.com/avocado-framework/aexpect/pull/106#issuecomment-1301842767 but I guess you insist on this

pevogam:
Hi @ldoktor may I ask for the reason to close the pull request?

ldoktor:
@pevogam could you please re-open this one for main? IIRC the only missing thing is the formatting and the verbosity (potentially merging multiple messages into one)

pevogam:
Hi @ldoktor, 

> @pevogam as I mentioned earlier, I don't mean to block this, I was just curious about the direction and scope. It seems to be quite ambitious and I think it'd make sense to get in touch with @clebergnu and @beraldoleal to avoid reinventing the wheel as they are also interested in extracting and generalizing the `avocado.utils`. In the mean-time as long as it's just a single file, let's keep this in `aexpect`, but the moment it becomes bigger let's remove it and make a separate `aexpect.ops` sub-project (could be in the same tree, no problem with that)

Alright then, let's consider this pull request as too platform/system specific and thus ambitious in its anticipated scope and I will drop it on my side too to keep the `aexpect.ops` module as command generic as possible. However, the one-file-only requirement you stated is too harsh and as already commented by @lmr and @clebergnu the threshold for "separate project" (independently of which git tree) is still too far. I would thus prefer to be able to at least contribute a separate ops module for remote aexpect sessions to windows platforms without triggering your acceptance limits if possible as the windows ops will also remain as generic as possible.

pevogam:
> Is the lazy logging formatting really obligatory? I am aware there are optimizations for it but there are also lots of counterargument against it and for using fstrings for logging messages instead. So in short, how hard set in stone is this requirement? I had to convert to lazy formatting for another pull request too but this now makes me explicitly ask about this.

I finally found my source regarding this:

https://github.com/PyCQA/pylint/issues/2354#issuecomment-414526879

In short there are overwhelming reasons why fstrings in log messages would make much sense too and why on our side we rather settled for this approach.

ldoktor:
> > Is the lazy logging formatting really obligatory? I am aware there are optimizations for it but there are also lots of counterargument against it and for using fstrings for logging messages instead. So in short, how hard set in stone is this requirement? I had to convert to lazy formatting for another pull request too but this now makes me explicitly ask about this.
> 
> I finally found my source regarding this:
> 
> [PyCQA/pylint#2354 (comment)](https://github.com/PyCQA/pylint/issues/2354#issuecomment-414526879)
> 
> In short there are overwhelming reasons why fstrings in log messages would make much sense too and why on our side we rather settled for this approach.

It's not 100% strict requirement and that comment makes good points. In aexpect we don't usually need to care about quickness except of few critical sections. I'm still more for the traditional way as in case of weird arguments we would get a nice message instead of just formatting error without details, but for usually-printed messages f-strings might be an option.

pevogam:
> > > Is the lazy logging formatting really obligatory? I am aware there are optimizations for it but there are also lots of counterargument against it and for using fstrings for logging messages instead. So in short, how hard set in stone is this requirement? I had to convert to lazy formatting for another pull request too but this now makes me explicitly ask about this.
> > 
> > 
> > I finally found my source regarding this:
> > [PyCQA/pylint#2354 (comment)](https://github.com/PyCQA/pylint/issues/2354#issuecomment-414526879)
> > In short there are overwhelming reasons why fstrings in log messages would make much sense too and why on our side we rather settled for this approach.
> 
> It's not 100% strict requirement and that comment makes good points. In aexpect we don't usually need to care about quickness except of few critical sections. I'm still more for the traditional way as in case of weird arguments we would get a nice message instead of just formatting error without details, but for usually-printed messages f-strings might be an option.

I guess the formatting error without details could be considered a temporary limitation of python regarding the still fairly newly introduced fstrings that could be lifted in the future. But I am glad I could finally point to the source of my otherwise not properly backed arguments above.

PR #108, opened 2022-11-03
Allow to calculate hashes on parts of files

pevogam:
This is especially useful for large files where calculating the hash can take long and even time out.

### DIFF
--- a/aexpect/ops.py
+++ b/aexpect/ops.py
@@ -424,13 +424,15 @@ def grep(session, expr, path, check=False, flags="-a"):
 ###############################################################################
 
 
-def hash_file(session, filename, method='md5'):
+def hash_file(session, filename, size="", method="md5"):
     """
     Calculate hash of given file from a session.
 
     :param session: session to run the command on
     :type session: ShellSession
     :param str filename: full path of file that should be hashed
+    :param str size: calculate hash from initial data chunk of some size,
+                     e.g. 1M, 1G, etc.
     :param str method: method for hashing
     :returns: hash of file, a hex number
     :rtype: str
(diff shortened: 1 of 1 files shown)


aexpect/ops.py
```diff
@@ -424,23 +424,29 @@ def grep(session, expr, path, check=False, flags="-a"):
 ###############################################################################
 
 
-def hash_file(session, filename, method='md5'):
+def hash_file(session, filename, method="md5", first_chunk=""):
```
ldoktor, line 427: Naming is always hard and I am not against "first_chunk" but what would you say about `size` argument instead?

pevogam: Sounds great and even more compatible with the `avocado.utils.crypto` API (`def hash_file(filename, size=None, algorithm="md5")` but do you mind if I also add size as a second argument to achieve full compliance? I know it is not advised to change the order of previously defined arguments but considering this is in early stages of usages perhaps I could still take the jump?

ldoktor: Sure, I do consider this module in alpha stage, still hammering out the APIs. Thanks for spotting this out and feel free to adopt the API from the `avocado.utils.crypto` module.

aexpect/ops.py
```diff
@@ -424,23 +424,29 @@ def grep(session, expr, path, check=False, flags="-a"):
...
+                     e.g. 1M, 1G, etc.
     :param str method: method for hashing
     :returns: hash of file, a hex number
     :rtype: str
     :raises: :py:class:`RuntimeError` if hash command failed (e.g. file not
              found) or resulting output does not have expected format
     """
     cmd = f"{method}sum"
+    if size:
+        cmd = f"head -c {size} {quote(filename)} | md5sum"
```
ldoktor, line 444: you should use `{cmd}` instead of the `md5sum`

pevogam: Hehe, silly me. Thanks for the nice catch!

ldoktor (approved):
Thank you, looks good.

PR #110, opened 2022-11-28
Codeql

ldoktor:
Enable LGTM directly on GH and fix the issues we want to address.

closes: #109

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -200,12 +200,15 @@ class Spawn:
                 pass
 
         # Open the reading pipes
-        try:
-            assert is_file_locked(self.lock_server_running_filename)
-            for reader, filename in self.reader_filenames.items():
-                self.reader_fds[reader] = os.open(filename, os.O_RDONLY)
-        except (AssertionError, OSError):
-            pass
+        if is_file_locked(self.lock_server_running_filename):
+            try:
+                for reader, filename in self.reader_filenames.items():
+                    self.reader_fds[reader] = os.open(filename, os.O_RDONLY)
+            except OSError:
+                LOG.warning("Failed to open reader '%s'", filename)
+        else:
+            LOG.warning("Not opening readers, lock_server_running "
+                        "not locked")
 
         # Allow the server to continue
         unlock_fd(lock_client_starting)
@@ -275,6 +278,7 @@ class Spawn:
             try:
                 os.close(fd_reader)
             except OSError:
+                # It was probably already closed
                 pass
 
     def _close_aexpect_helper(self):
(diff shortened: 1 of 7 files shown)


aexpect/client.py
```diff
@@ -415,7 +416,7 @@
             os.write(proc_input_pipe, cont.encode(self.encoding))
             os.close(proc_input_pipe)
         except OSError:
-            pass
+            LOG.warning("Failed to send '%s'", cont)
```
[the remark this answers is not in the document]
lmr: Interesting warning. I don't think it's too important right now but we might want to write a scrubber of strings such as passwords so that they are not logged.

lmr (approved):
Good fixes @ldoktor, they LGTM!

aexpect/client.py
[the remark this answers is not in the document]
ldoktor: Given the purpose of this tool it's IMO important to log the plaintext password.

ldoktor:
Thank you, Lucas, for review, I added extra commit to treat null/empty output_params, which was polluting my logs.

aexpect/client.py
```diff
@@ -205,7 +205,7 @@ def __init__(self, command=None, a_id=None, auto_close=False, echo=False,
             for reader, filename in self.reader_filenames.items():
                 self.reader_fds[reader] = os.open(filename, os.O_RDONLY)
         except (AssertionError, OSError):
-            pass
+            LOG.warning("Failed to open reader '%s'", filename)
```
clebergnu, line 208: AFAICT this warning message would be appropriate for lines 205-206, but not for 204.  If that's the case, I'd suggest coming up with either a generic message or two different expect blocks that can give a more precise warning (preferred).

scripts/aexpect_helper
```diff
@@ -130,7 +130,6 @@ def main():     # too-many-* pylint:disable=R0914,R0912,R0915
 
             # Read from child and write to files/pipes
             server_log.info('Entering main read loop')
-            status = 1  # lgtm [py/multiple-definition]
```
clebergnu, line 133: OK, so line 183 will only be reached when a valid `status` is set on line 172. LGTM.

clebergnu (commented):
Hi @ldoktor , great work here!  I've left a comment/question about the warning messages.  Other than that, it LGTM.

aexpect/client.py
ldoktor: good catch

ldoktor:
Changes:

Split the lock/open warning messages

clebergnu (approved):
LGTM, thanks!

PR #111, opened 2022-12-12
Implement a reusable subcommand output and status processing routine

pevogam:
As most calls would check for nonzero exit status and strip the command output let's reuse the same logic as much as we can.

### DIFF
--- a/aexpect/ops.py
+++ b/aexpect/ops.py
@@ -60,6 +60,14 @@ from aexpect.exceptions import ShellCmdError
 LOG = logging.getLogger(__name__)
 
 
+def _process_status_output(command, status, output):
+    if output != "":
+        output = output.strip()
+    if status != 0:
+        raise ShellCmdError(command, status, output)
+    return status, output
+
+
 ###############################################################################
 # stat(2)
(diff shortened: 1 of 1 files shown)


pevogam:
The failed version here is specifically for Python 3.6 due to unavailable platform in the CI.

lmr (approved):
LGTM, thank you!

aexpect/ops.py
```diff
@@ -60,6 +60,14 @@
 LOG = logging.getLogger(__name__)
 
 
+def _process_status_output(command, status, output):
+    if output != "":
+        output = output.strip()
+    if status != 0:
+        raise ShellCmdError(command, status, output)
+    return status, output
```
ldoktor, line 68: Given the status will always be 0, do you think it is necessary to report it back?

pevogam: The idea here is simply to have a consistent interface with a "brief pass over the status and output" where some common processing could raise errors and mutate any of the two argument. It is true only some such processing is done right now but I think this could more easily generalize would we need further processing of both.

PR #112, opened 2022-12-12
Add a generic ops module for windows platforms

pevogam:

### DIFF
--- a/aexpect/ops_windows.py
+++ b/aexpect/ops_windows.py
@@ -1,0 +1,578 @@
+# This Python file uses the following encoding: utf-8
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Intra2net AG and aexpect contributors
+# Authors : Samir Aguiar <samir.aguiar@intra2net.com>
+
+"""
+A list of frequent operations performed through remote sessions to windows platforms.
+
+PowerShell:
+A section for PowerShell-based commands and scripts.
+
+registry ops:
+Registry key manipulation.
+
+filesystem ops:
+Operations concerning files on Windows e.g. file paths and directories.
+
+process ops:
+Operations concerning process management e.g. start, kill, and run as.
+
+network ops:
+Network configuration and downloads.
+"""
+
+import re
+import uuid
+import logging
+from textwrap import dedent
+from base64 import b64encode
+from enum import Enum, auto
+
+# avocado imports
+from aexpect.exceptions import ShellError, ShellProcessTerminatedError
+
+LOG = logging.getLogger(__name__)
+
+
+###############################################################################
+# PowerShell
+###############################################################################
+
+
+def ps_cmd(session, script, timeout=60):
+    """
+    Run one or multiple PowerShell commands.
+
+    :param session: session of the Windows guest
+    :type session: :py:class:`aexpect.client.RemoteSession`
+    :param str script: one or multiple PowerShell commands
+    :param int timeout: seconds to wait for the command to complete
+    :returns: the output of the command
+    :rtype: str
+
+    The string with commands is encoded in base64, which means that
+    a whole multiline script can be passed without problems.
+
+    This function was slightly based on a similar function from `pywinrm`.
+    """
+    def nicely_log_str(header, string):
+        LOG.info(header)
+        LOG.info("-" * len(header))
+        for line in string.splitlines():
+            LOG.info(line)
+        LOG.info("\n")
+
+    nicely_log_str("Running PowerShell script", dedent(script))
+    # must use utf16 little endian on Windows
+    encoded_cmd = b64encode(script.encode("utf_16_le")).decode("ascii")
+    try:
+        output = session.cmd(f"powershell -NoLogo -NonInteractive -OutputFormat text "
+                             f"-ExecutionPolicy Bypass -EncodedCommand {encoded_cmd}",
+                             timeout=timeout)
+
+        # there was an error but the exit code was still zero
+        if output.startswith("#< CLIXML") and "<S S=\"Error\">" in output:
+            raise ShellError(script, output)
+
+        # TODO: PowerShell sometimes ignores the -OutputFormat and outputs
+        # as XML. This has been fixed on newer versions but we have yet to update:
+        # https://github.com/PowerShell/PowerShell/pull/8115
+        if output.startswith("#< CLIXML") and "<Objs Version" in output:
+            output = "\n".join(output.splitlines()[1:-1])
+
+        nicely_log_str("PowerShell script output", output)
+        return output.strip()
+    except ShellError as error:
+        # `cmd` is encoded here, replace before raising
+        error.cmd = script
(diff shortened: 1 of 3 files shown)


pevogam:
I will take a look at the more or less minor CI incompatibilities of style soon but first want to get your overall opinions of which there are likely many ^^

aexpect/ops_windows.py
```diff
@@ -0,0 +1,584 @@
...
+
+    :param session: session of the Windows guest
+    :type session: :py:class:`aexpect.client.RemoteSession`
+    :param str script: one or multiple PowerShell commands
+    :param int timeout: seconds to wait for the command to complete
+    :returns: the output of the command
+    :rtype: str
+
+    The string with commands is encoded in base64, which means that
+    a whole multiline script can be passed without problems.
```
lmr, line 65: Neat

aexpect/ops_windows.py
```diff
@@ -0,0 +1,578 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: Intra2net AG and aexpect contributors
+# Authors : Samir Aguiar <samir.aguiar@intra2net.com>
+
+"""
+A list of frequent operations performed through remote sessions to linux platforms.
```
ldoktor, line 18: Shouldn't this be `windows` platforms?

ldoktor:
Hello @pevogam, what a nice set of utilities. In general (apart from the this-should-be-an-optional-plugin) I like the addition of windows utils. Have you considered re-using the same API linux version uses? I think it'd simplify the usage for people who deal with both simultaneously. Obviously they would have to still change the values (`-h` vs `/h`, ...) but overall the experience might be better.

Even better would be to add a helper `ops` class to work with arbitrary session, providing (detecting and/or caching) the system type. IIRC virttest did that for some of the libs and it was very convenient and helped keeping the APIs the same for both OSes.

Please, let me know about the APIs, I'd welcome that but given I'm not the target audience I don't have a strong opinion about it. By any chance, do you know of a simple way to test it out without windows? (eg. via wine?) I could probably setup a VM but honestly I'd rather just review the code without functional testing to spend a day of configuring something just for one PR (per a few months).

aexpect/ops_windows.py
```diff
@@ -0,0 +1,578 @@
...
+    session.close()
+    LOG.info("Session successfully killed")
+
+
+###############################################################################
+# network ops
+###############################################################################
+
+
+def find_unused_port(session, start_port=10024):
```
ldoktor, line 496: Is this a typo or are ports 10024+ special in windows?

aexpect/ops_windows.py
pevogam: Nice catch and a quick copy paste on my side, thanks!

aexpect/ops_windows.py
pevogam: The main reason is that typical ports used by Avocado VT for windows vms are:
```
        shell_client = nc
        shell_port = 10022
        file_transfer_client = rss
        file_transfer_port = 10023
```
but I guess we could also set the default to something more standard outside of this scope. Do you have any specific suggestions?

pevogam:
> Hello @pevogam, what a nice set of utilities.

All credit goes to @samiraguiar who wrote most of the code of this module years ago :clap: 

> In general (apart from the this-should-be-an-optional-plugin) I like the addition of windows utils. Have you considered re-using the same API linux version uses? I think it'd simplify the usage for people who deal with both simultaneously. Obviously they would have to still change the values (`-h` vs `/h`, ...) but overall the experience might be better.
> 
> Even better would be to add a helper `ops` class to work with arbitrary session, providing (detecting and/or caching) the system type. IIRC virttest did that for some of the libs and it was very convenient and helped keeping the APIs the same for both OSes.

Absolutely! Sounds like the best way forward and if you confirm I could create an issue about it to track progress in such unification and connect pull requests with this issue in the future. How does that sound?

> Please, let me know about the APIs, I'd welcome that but given I'm not the target audience I don't have a strong opinion about it. By any chance, do you know of a simple way to test it out without windows? (eg. via wine?) I could probably setup a VM but honestly I'd rather just review the code without functional testing to spend a day of configuring something just for one PR (per a few months).

Not sure if this would make your life genuinely easier but all utilities have been integration tested and are fully used in our avocado-based test suite which includes quite many windows vms (also of different versions). As most of these utilities are practically a "test boundary" (destructive environment interacting operations wrapped in a bit of code) the only appropriate testing I would recommend here is integration testing which would need corresponding environments. If you are interested in the future we could use avocado(-vt) for integration testing the aexpect codebase too. This is what we do but indirectly via test suite focused on other products.

ldoktor (approved):
Hello @pevogam, I haven't tested this on windows, but I overlooked it and tried some on `wine cmd`. Most of it did not work because `wine cmd` does not support much, but overall it looks like a decen alpha version.

Please consider the merging the API for commands that are similar to `ops_linux` and ideally creating `ops` to abstract them, but it's really just a suggestion. I am not the primary user so it's your call.

Please let me know if you want me to merge this or if you still continue on developing this.

pevogam:
Hi @ldoktor as all thing go gradual I would suggest we merge this so that it becomes available for use and I can then create and track (on my side but publicly) an issue regarding a potential merge. Perhaps it will need some reevaluation first though because there are lots of specialized commands that are only available on linux (x)or windows. Even if the user can abstract away from a few shared concepts in each, they will sooner or later in their use most likely end up needing something particular for their platform and thus the transparency will be replaced by an inevitable need to differentiate among them and import only some. These are just some thoughts that I had after the last comments we had  here that might imply that it is better to use the utilities more widely first before deciding for such an action. What do you think?

ldoktor:
That is exactly what I had in mind. Given those modules are really an alpha (beta in best case) stage, I'm fine with merging this as is to allow upstream cooperation.

ldoktor:
Oh, could you please add the `pylint: disable=W0511` to the lines 89 and 558 so the CI passes? (or globally for the whole file)

pevogam:
> pylint: disable=W0511

I was actually wondering about this: what is the right way accepted by Aexpect to write in-line todo-s? Is it really FIXME and alice-in-wonderland stuff like this? Is this enforceable? I can also adjust the todos or clarify this first so that we take a better approach than simply locally disabling.

ldoktor:
Hello @pevogam, well a while ago we made the experiment to clean code according to pylint. 100% of it. This is a remnant from that era and, as we cleaned all TODO/FIXME/XXX/... it was not needed. The theory behind is that production software should not contain TODOs, it should be finalized before it's delivered (so while it lives in staging, other repo, other module, ...). The question is, whether it's right for this project. I am open to changing this (therefore adding the `W0511` to `make check` target)

pevogam:
> Hello @pevogam, well a while ago we made the experiment to clean code according to pylint. 100% of it. This is a remnant from that era and, as we cleaned all TODO/FIXME/XXX/... it was not needed. The theory behind is that production software should not contain TODOs, it should be finalized before it's delivered (so while it lives in staging, other repo, other module, ...). The question is, whether it's right for this project. I am open to changing this (therefore adding the `W0511` to `make check` target)

Hi @ldoktor and thanks for the explanation. From what I gather it doesn't make much sense for us to store these TODO-s as GH issues and I take code-local TODO-s simply as a lower priority consider-when-you-refactor-this-code situations. In that sense I am not sure about the reasoning that TODOs should be forbidden within the code as some of them could definitely benefit from localization and remains secondary at the same time if e.g. they indicate direction of development for a particular implementation which doesn't make much sense to discuss in the overall system scope. In this sense I would prefer to just add the flag to the disabled ones and will push with this change included but feel free to block the merge if you disagree or I didn't understand your intent correctly.

PR #113, opened 2023-01-17
Add remote linux glob behavior to expand paths from patterns

pevogam:
This was previously done via `ops.ls` but the `ls` would return a "serious error" exit code 2 when the expanded glob pattern results in zero paths. The current reimplementation makes use of `echo` to achieve the same with safe zero path outcome that won't hide more serious errors.

### DIFF
--- a/aexpect/ops_linux.py
+++ b/aexpect/ops_linux.py
@@ -237,6 +237,30 @@ def ls(session, path, quote_path=True, flags="-1UNq"):  # pylint: disable=C0103
     return output.splitlines()
 
 
+def glob(session, glob_pattern):
+    """
+    Find files matching given pattern through a session.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str glob_pattern: pattern of filenames where `*`, `?` and such match
+                             any string or not-empty-string and so on.
+    :returns: file names matching pattern, including path (like glob.glob)
+    :rtype: [str]
+
+    The current remote implementation has the limitation that it would hide
+    any `find` exit status 1 that might not be related to the glob expansion.
+    Alternative implementations are either shell specific or use `echo` and
+    thus cannot handle spaces in paths well.
+    """
+    cmd = f'find {glob_pattern} -maxdepth 0'
+    status, output = session.cmd_status_output(cmd)
+    if status == 1:
+        return []
+    status, output = _process_status_output(cmd, status, output)
+    return output.splitlines()
+
+
 def move(session, source, target, quote_path=True, flags=""):
     """
     Move file or directory from source to target / rename source to target.


pevogam:
@ldoktor Do you think it is also possible to release 1.6.5 after this pull request is merged? This would greatly simplify deploying all the merged work from the past weeks / years on our side and would be really appreciated.

pevogam:
@lmr Could you help review this and potentially release?

aexpect/ops_linux.py
```diff
@@ -216,6 +216,29 @@ def is_nonzero_size_file(session, path):
...
+    :returns: file names matching pattern, including path (like glob.glob)
+    :rtype: [str]
+
+    Much like :py:func:`glob.glob` but implemented using `echo`.
+    """
+    cmd = f'echo {glob_pattern}'
+    status, output = session.cmd_status_output(cmd)
+    status, output = _process_status_output(cmd, status, output)
+    paths = []
+    for path in output.split():
```
ldoktor, line 236: This will not work for paths containing spaces. (eg `touch "a b"`)

aexpect/ops_linux.py
```diff
@@ -216,6 +216,29 @@ def is_nonzero_size_file(session, path):
...
+        if test(session, path, "-e"):
```
ldoktor, line 237: I liked this idea to ensure the path exists, but the cons is it's really slow. For 100 files it took 50s on my PC. Provided you fix the glob to really return the correct paths, this should not be needed, right?

aexpect/ops_linux.py
pevogam: Hmmm this is indeed a limitation however, I would prefer to simply put a disclaimer in the docstring as there is no better solution I would prefer. As you can see in our original discussion the only other option is shell specific and I would prefer not to go in this direction. We already ruled out the `ls` hiding status and this is the only option that remains. Is this ok?

aexpect/ops_linux.py
pevogam: We could then just fall back to the original `ls` solution hiding the status. Even though I had to investigate for solutions for not hiding the status and we end up back where we started which sounds like some wasted time and effort. Or do you have a better solution you could propose?

aexpect/ops_linux.py
ldoktor: I'd probably use `MATCHES=({globpattern})\nfor MATCH in "${MATCHES[@]}"; do echo "$MATCH"; done` although I'm not 100% sure we won't end up with some weird chars issues as well.

pevogam: This sounds pretty much like my comment https://github.com/avocado-framework/aexpect/pull/98#discussion_r988658483 which remains shell specific or am I missing something?

aexpect/ops_linux.py
ldoktor: Looks like there is no perfect solution. There will always be some limitation. The `ls` solution should do in most cases s I think it makes sense to use that one and leave it up to people to use custom handling in case this fails.

pevogam: Alright, pushed the changes just now, I think CI will pass.

aexpect/ops_linux.py
```diff
@@ -237,6 +237,30 @@ def ls(session, path, quote_path=True, flags="-1UNq"):  # pylint: disable=C0103
...
+    any `ls` exit status 2 that is a serious error not related to the glob
+    expansion. Alternative implementations are either shell specific or use
+    `echo` and thus cannot handle spaces in paths well.
+    """
+    cmd = f'ls {glob_pattern}'
+    status, output = session.cmd_status_output(cmd)
+    if status == 2:
+        return []
+    status, output = _process_status_output(cmd, status, output)
+    return output.splitlines()
```
ldoktor, line 261: In order to be able to split by lines it's necessary to enforce `-1` argument:

```
>>> aexpect.ops_linux.ls(a, "*", False, '-1UNq')
['a', 'a b']
>>> aexpect.ops_linux.ls(a, "*", False, '')
["a  'a b'"]
```

Also note that the `ls` approach can result in surprising output when in use with dirs:

```
>>> aexpect.ops_linux.ls(a, "/tmp/b", False, '-1')
['a', 'b']
>>> aexpect.ops_linux.ls(a, "/tmp/b/*", False, '-1')
['/tmp/b/a:', '', '/tmp/b/b:']
```

I don't want to draw the line here, it depends on which features you, as the primary user, need. I'm fine with any approach as long as it gets properly documented and is somehow systematic. So I'd suggest the use of glob and arrays, or ls with the usual arguments (perhaps just call the `ops_linux.ls` and catch for status 2?)

pevogam: All good points and integration tests here are still running hence I didn't notice this yet. How about using something that will simplify all of these residues of user formatted output like `find`?
```
[pevogam@drogon androids]$ find c*
c
c b
[pevogam@drogon androids]$ ls c*
'c b'

c:
```
I guess it depends what the user would expect to get under glob but I would reckon receiving a guaranteed list of paths seems much more reasonable expectation than dir-ls-like behavior.

pevogam: (the `-1` was indeed important but my testing didn't catch it because of some default `ls` settings on some linux distros)

ldoktor: I though about find for a while but decided not to pursuit it further, but it can also work. Only I'd suggest using `-maxdepth 0`. With that it works well for all of the variants I tried.

ldoktor: find /tmp/*/*a* -maxdepth 0

ldoktor: find /tmp/b/* -maxdepth 0

pevogam: Thanks and I apologize for the micromanagement here - pushed and CI passing, will need more time to confirm the integration tests we have but I think we are both feeling confident about the final solution now.

ldoktor (approved):
Thanks, seems to work well now

aexpect/ops_linux.py
ldoktor: No problem, that is what reviews are for.

pevogam:
> @ldoktor Do you think it is also possible to release 1.6.5 after this pull request is merged? This would greatly simplify deploying all the merged work from the past weeks / years on our side and would be really appreciated.

Thanks a lot @ldoktor! What are your plans regarding future release of aexpect? Do you think something like 1.6.5 will be possible anytime in the near future?

ldoktor:
I don't have the rights to release aexpect, so the decision is on @lmr, anyway I'd appreciate to discuss the https://github.com/avocado-framework/aexpect/pull/114 first as this would be the first official release with the new libraries. So Lucas, would you please take a look a that?

PR #115, opened 2023-01-23
 RFC: Split optional modules

ldoktor:
the aexpect package is becoming bigger, it might be useful to split it across multiple optional modules. Let's start with "remote" and "ops" modules. The difference is only in way we import those modules, instead of:

we should do:

    from aexpect_remote import remote_door

although, for compatibility reasons, some of the modules are still available directly in aexpect, when installed on the system.


ldoktor:
This was accidentally pushed and reverted from upstream, the original issue along with some discussion is here: https://github.com/avocado-framework/aexpect/pull/114

lmr:
So, my opinions on the subject:

1. Yes, the library has increased in size, but it is not so large that it would be necessary to split
2. Since the new functionality introduced by @pevogam has been merged, it makes sense to release it as is
3. Looking at the current code, I see it could benefit from using namespaces a bit more - `ops` and `remote` should become separate directories under `aexpect`, same as we have for `utils`. I believe this would make it easier to create separate RPM packages out of the source code.
4. The problem with 3. is that this could introduce some API incompatible changes to the code, which would force us to handle updates for all the aexpect dependencies. I would like to gauge people's interest on doing that.

So I would say, release what is there per 2., and let's start to consider the restructuring proposed in 3.

lmr:
We have a difference of opinion on the merits of having all code shipped together or separating it inside the aexpect code.

I believe I have a proposal that could make up for a good compromise between the 2 views:

Create a new library with the higher level constructs that depends on aexpect and host it under the avocado-framework area. @pevogam would be the maintainer, as he contributed most of the code available in it and we can keep aexpect small and self contained. How about that?

pevogam:
I think I made a mistake and posted my answers in #114 which was merged and then a new one (this pull request) was opened in its place, I glossed over that and kept posting in the original pull request. So in short: all my arguments are placed in #114.

ldoktor:
Let's close this as we decided not to split yet: https://github.com/avocado-framework/aexpect/pull/114

PR #117, opened 2023-02-08
Avoid printing line when output_func not set

ldoktor:
On two places we used to call _print_line even though self.output_func was not set, which resulted in warnings. It would be nicer to put this check directly to the _print_line function, but it is used inside performance critical section so let's check that directly on the corresponding places.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -675,10 +675,11 @@ class Tail(Spawn):
                 else:
                     # No output is available right now; flush the bfr
                     if bfr:
-                        _print_line(bfr)
+                        if self.output_func:
+                            _print_line(bfr)
                         bfr = ""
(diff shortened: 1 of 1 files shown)


pevogam:
Will run some testing with this change, let's see.

ldoktor:
@pevogam did this help, shell we merge this before the release?

pevogam:
Hi @ldoktor unfortunately not and a comparison with earlier versions revealed we have had the issue for a long time too, it simply sunk in the avocado logs. But at the same time this change seems harmless so if you are confident about it (we didn't get any problems with it either from what I can tell) you can also go ahead and merge.

ldoktor:
OK, it should decrease the amount of false-warnings, so let's get this in.

PR #119, opened 2023-02-28
python-aexpect.spec: Update spec with latest version

lmr:
I forgot to make that update when cutting 1.7.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PYTHON_DEVELOP_ARGS=$(shell if ($(PYTHON) setup.py develop --help 2>/dev/null | 
 DESTDIR=/
 BUILDIR=$(CURDIR)/debian/aexpect
 PROJECT=aexpect
-VERSION="1.6.4"
+VERSION="1.7.0"
 COMMIT=$(shell git log --pretty=format:'%H' -n 1)
 SHORT_COMMIT=$(shell git log --pretty=format:'%h' -n 1)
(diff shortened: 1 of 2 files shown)


lmr:
@clebergnu when you have some time, please update the Fedora packages. I have been doing some package maintenance for Amazon Linux, so I can also help with maintenance in case you want.

ldoktor (changes requested):
Hello Lucas, there is one more file to be changed:

```patch
diff --git a/Makefile b/Makefile
index 6c2dc73..2fe0ea5 100644
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PYTHON_DEVELOP_ARGS=$(shell if ($(PYTHON) setup.py develop --help 2>/dev/null |
 DESTDIR=/
 BUILDIR=$(CURDIR)/debian/aexpect
 PROJECT=aexpect
-VERSION="1.6.4"
+VERSION="1.7.0"
 COMMIT=$(shell git log --pretty=format:'%H' -n 1)
 SHORT_COMMIT=$(shell git log --pretty=format:'%h' -n 1)
 COMMIT_DATE=$(shell git log --pretty='format:%cd' --date='format:%Y%m%d' -n 1)
```

ldoktor (approved):
Thank you, seems to work well although I'm wondering how the `make rpm-release` will work (maybe a new tag will be required, but IIRC Fedora uses different targets so perhaps it's not necessary; let's wait for @clebergnu or @mmathesius)

Issue #116, opened 2023-02-07
Printing IP or hostname with windows session fails


pevogam:
There is some problem with windows sessions when using the session prefix like:
```
[stdlog] 2023-02-07 13:40:58,382 aexpect.client WARNI| Failed to print_line '()' '192.168.51.1Please wait...'
[stdlog] 2023-02-07 13:40:58,382 aexpect.client WARNI| Failed to print_line '()' '192.168.51.1Microsoft Windows [Version 10.0.14393]'
[stdlog] 2023-02-07 13:40:58,382 aexpect.client WARNI| Failed to print_line '()' '192.168.51.1(c) 2016 Microsoft Corporation. Alle Rechte vorbehalten.'
[stdlog] 2023-02-07 13:40:58,382 aexpect.client WARNI| Failed to print_line '()' '192.168.51.1'
```
I will investigate this shortly but after recent changes of avocado removed the logged line in the format I will have to run in a special older environment first.

ldoktor:
Hello Plamen, this is happening when the `self.output_func` is not set and we flush the data on line `678` or `682`. I meant to look at that but got more important matters.

ldoktor:
(alternatively it might happen when we have `self.output_func` set when we get to line 668 and then removed before we run line 635 or 633)

ldoktor:
Note these were there all the time, they were just silently ignored. IIRC I wanted to leave it on warning for now to evaluate later. Would you have any preference? The `_print_line` function is kind-of performance intensive so I'd prefer not putting there yet another check for `self.output_func`, but perhaps we could only warn when `self.output_params` are used plus also add the `if self.output_func` to the lines 677 and 681.

ldoktor:
https://github.com/avocado-framework/aexpect/pull/117 this tackles the extra places where we ought to check `self.output_func` presence, I have tested that and it made a huge difference in my workflow (lots of previously neglected warnings are gone). As I mentioned earlier, we could still get the warning in case the user changes `self.output_func`, but fixing that would require locking and I don't think the benefit would be so big (functionally it would be fine, only a warning might appear when `self.output_func` is removed).

pevogam:
@ldoktor Since so far we did what we could here and this turned out to be a false alarm of something already known it is better to close it. Let's see if later redesigns might help us improve on this further when the right time comes.

PaulYuuu:
Hello @pevogam, I also hit this problem, and after investigation, I think the following code should be updated.
[...]

PR #120, opened 2023-03-09
remote_login: Fix problem of losing logs before setting output_func

PaulYuuu:
After RemoteSession initialization and before set_output_func, There are very short time intervals, but sometimes messages will be lost or printed via `print_line`. The solution is to define `set_output_func`/`set_output_params` when initializing RemoteSession because it natively supports them.

Signed-off-by: Yihuang Yu <yihyu@redhat.com>

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -411,22 +411,24 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
         cmd = f"nc {' '.join(extra_params)} {host} {port}"
     else:
         raise LoginBadClientError(client)
+    output_params = ()
+    if log_filename:
+        output_params = (log_filename,)
+        log_filename = os.path.basename(log_filename)
 
     if verbose:
         LOG.debug("Login command: '%s'", cmd)
-    session = RemoteSession(cmd, linesep=linesep, output_prefix=host,
+    session = RemoteSession(cmd, linesep=linesep, output_func=log_function,
+                            output_params=output_params, output_prefix=host,
                             prompt=prompt, status_test_command=status_test_command,
                             client=client, host=host, port=port,
                             username=username, password=password)
+    session.set_log_file(log_filename)
     try:
         handle_prompts(session, username, password, prompt, timeout)
     except Exception:
         session.close()
         raise
-    if log_filename:
-        session.set_output_func(log_function)
-        session.set_output_params((log_filename,))
-        session.set_log_file(os.path.basename(log_filename))
     return session
 
 


aexpect/remote.py
```diff
@@ -423,10 +428,7 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     except Exception:
         session.close()
         raise
-    if log_filename:
-        session.set_output_func(log_function)
-        session.set_output_params((log_filename,))
-        session.set_log_file(os.path.basename(log_filename))
+    session.set_log_file(log_filename)
```
PaulYuuu, line 431: About this line, it needs to be set after the initialization, I don't find a better way to update it. But by default, session.log_file is `None`, so even if `log_filename` is `None`, this is just a fake operation and will not affect the session.

PaulYuuu:
I think this should fix issue #116.

To be honest, there has the easiest way to workaround way provided by @luckyh, is switching the order of `set_output_func` and `set_output_params`, this can address the `print_line` warning, but we will still lose some log messages.
```python
    if log_filename:
        session.set_output_params((log_filename,))
        session.set_output_func(log_function)
        session.set_log_file(os.path.basename(log_filename))
```

If we don't want to refactor the remote_login part, I would pick this one. @pevogam @ldoktor, do you have some suggestions about this patch?

ldoktor (commented):
Overall I do like this better, setting the function on init should be safer. Please consider moving the `set_log_file` directly after init so the `log_file` gets cleared on `handle_prompts` failure.

Btw: Looking at the code now, the original issue was in the order of calling `set_output_func` and `set_output_params`. The `set_output_params` should be called first as the `set_output_func` creates the thread which needs the `output_param` to be set already. Anyway all of this should be properly handled by using the init arguments...

ldoktor:
Oh, @PaulYuuu you've already noticed that earlier so kudos to you. Anyway using the init arguments should be the safest way as we have to make sure the order is correct on such occasions. So all in all I do support this PR, and before merging it I'd like you to re-consider where the `set_log_file` is executed.

aexpect/remote.py
PaulYuuu: Yes, your suggestion is safer, I will update it in the next revision.

PaulYuuu:
> Oh, @PaulYuuu you've already noticed that earlier so kudos to you. Anyway using the init arguments should be the safest way as we have to make sure the order is correct on such occasions. So all in all I do support this PR, and before merging it I'd like you to re-consider where the `set_log_file` is executed.

Thanks @ldoktor for your detailed comments. I have moved `set_log_file` to a better place. Actually I also consider setting log_file when initializing `RemoteSession`, but this needs to modify lots of code, so let's keep the current one. Please help to review if this is fine for you.

pevogam (approved):
I ran a minimal test set with these changes and they look good :heavy_check_mark: 

I only wonder if the argument about "too many places to change" or "possible API to break" really takes precedence in also setting the `log_filename` from within the remote session's constructor (or even changing the constructor to allow for a single `output_params` dictionary of options to be passed for all of `output_prefix`, `output_params`, `output_func`, etc.). The problem in this argument is that we take shortcuts with guaranteed short term gains but long term headache as we try to be backwards compatible.

In any case, the current changes might not go the full way but definitely make a step in this direction so perhaps this is good enough for now.

ldoktor (approved):
Yep, looks good to me. Thank you.

As for the API, yes, these changes are hard as backwards compatibility is a pain, but also a commitment. It's not just about adding yet another argument as if we wanted to do it properly, that would mean ideally removing bunch of init arguments and moving to better post-setup.

pevogam:
> Yep, looks good to me. Thank you.
> 
> As for the API, yes, these changes are hard as backwards compatibility is a pain, but also a commitment. It's not just about adding yet another argument as if we wanted to do it properly, that would mean ideally removing bunch of init arguments and moving to better post-setup.

Indeed, it needs more work - just let's keep in mind that finding workarounds now has larger complexity cost further down the road which was my point. Hopefully we can reduce it back before some major version release instead of offering indefinite backwards compatibility.

ldoktor:
Sure, if someone finds the time to do a complete overhaul, then yes. But changing the API per-feature IMO does not justify the breakage.

pevogam:
> Sure, if someone finds the time to do a complete overhaul, then yes. But changing the API per-feature IMO does not justify the breakage.

Agreed with that.

PR #121, opened 2023-03-21
Introduce async alternatives in the client command hierarchy

pevogam:
TODO: Consider best approach here and in the topmost nonblocking reading.

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -27,6 +27,7 @@ import select
 import subprocess
 import locale
 import logging
+import asyncio
 
 from aexpect.exceptions import ExpectError
 from aexpect.exceptions import ExpectProcessTerminatedError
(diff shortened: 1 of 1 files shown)


pevogam:
@ldoktor This is just a very drafty idea and I am just curious about your opinion in general and for alternatives in case there are obvious one I have definitely missed.

ldoktor (commented):
Hello @pevogam, it looks nice, I wouldn't be oppose to this, but I wouldn't benefit of it either. I played with asyncio briefly a while ago and found out it's not useful for tasks I'm using aexpect for (I liked it's usage in interactive games, looked clear there, but with regards to testing it seems too strict to not-really-educated me)

So my main question would be, do you plan to or know someone who would be using this API and would be willing to address potential issues?

@lmr @clebergnu what are your opinions?

pevogam:
> Hello @pevogam, it looks nice, I wouldn't be oppose to this, but I wouldn't benefit of it either. I played with asyncio briefly a while ago and found out it's not useful for tasks I'm using aexpect for (I liked it's usage in interactive games, looked clear there, but with regards to testing it seems too strict to not-really-educated me)
> 
> So my main question would be, do you plan to or know someone who would be using this API and would be willing to address potential issues?

Indeed there is a background for the idea - a remote test spawner introduced in https://github.com/avocado-framework/avocado/pull/5621 that makes use of aexpect. There are however ways to do it also without aexpect having asyncio capabilities but you can see the earlier commits there for an example.

> @lmr @clebergnu what are your opinions?

Feel free to compare with the pull request too regarding use cases for some source of details.

ldoktor:
@pevogam are there any plans about asyncio for aexpect or did you used a different approach?

pevogam:
> There are however ways to do it also without aexpect having asyncio capabilities but you can see the earlier commits there for an example.

Hi @ldoktor indeed as mentioned here we just opted for asyncio break points on different locations and at present our maintenance efforts will be much simpler if we do not add such capabilities. Let's revisit such a topic if more people become interested in it and can provide further arguments for it.

pevogam:
Of course, just in case there is something useful I still have these branches around:

https://github.com/pevogam/aexpect/tree/async-client
https://github.com/pevogam/aexpect/tree/asyncio-remote-copy

Anyone still interested feel free to check them out.

PR #123, opened 2023-05-08
Remote door tweaks

pevogam:
Adjust remote door file transfer port in case of custom choices and shorten remote door's control file output to a readable length both of which are small tweaks to make the remote door slightly more flexible.

These changes also allow us to run avocado tests with a remote spawner and perform distributed testing.


ldoktor (approved):
OK, makes sense to me. I'd prefer raising the full error, though, but I'm not the primary user and you log the full error before, so fine by me.

pevogam:
Thanks! Note that we still log the complete error via an error level logging message, we just abbreviate it for the final thrown error which often ends up in a test status or other circumstance where one would expect a short error message.

PR #126, opened 2023-12-08
requirements: Add setuptools

ldoktor:
the setuptools are no longer defaults and need to be stated explicitly.

### DIFF
--- a/aexpect/ops_windows.py
+++ b/aexpect/ops_windows.py
@@ -503,6 +503,7 @@ def find_unused_port(session, start_port=10024):
     :returns: unused port that was found
     :rtype: int
     """
+    # pylint: disable=C0301
     output = ps_cmd(session, f"""
         $port = {start_port}
         while($true) {{
@@ -521,6 +522,7 @@ def find_unused_port(session, start_port=10024):
 
         Write-Host "::unused=$port"
     """)
+    # pylint: enable=C0301
     port = re.search(r"::unused=(\d+)", output).group(1)
     return int(port)
 
--- a/requirements-travis.txt
+++ b/requirements-travis.txt
@@ -1,3 +1,3 @@
 setuptools==69.0.2
-pycodestyle==2.4.0
+pycodestyle==2.11.1
 inspektor==0.5.3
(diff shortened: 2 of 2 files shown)


ldoktor:
@pevogam could you please review this simple CI fix?

aexpect/ops_windows.py
```diff
@@ -503,6 +503,7 @@ def find_unused_port(session, start_port=10024):
     :returns: unused port that was found
     :rtype: int
     """
+    # pylint: disable=C0301
```
pevogam, line 506: I wonder here (and the commit message adding these lines doesn't answer it) why are these changes now? Is it because the CI didn't run for a while? At first I thought it is related to bumping some new version of linter but pydocstyle doesn't seem related here. Even if it was an offline CI (we could clarify this in the commit message) I still don't quite understand how these changes would pass undetected when the pull request was merged (I think the CI worked fine there).

pevogam (commented):
I have some unanswered questions from the intent of the commit messages but other than that thanks for restoring the CI here, a very vital for the health of the project.

aexpect/ops_windows.py
ldoktor: We are using the latest python3, which happens to be 3.12 now and it does not work with the old bumped pylint. So yes,I had to update the pylint and related modules and these seem to complain about this line (although it should have been probably detected earlier)

ldoktor:
Thank you, Plamen, for reviewing this, let me merge it so the CI works well again.

aexpect/ops_windows.py
pevogam: Ok, then let's prioritize restoring the CI. On my side I believe it is useful to investigate such external changes and document them in commit messages since pylint is not supposed to change its requirements on pre-existing and already linted issues like C0301.

ldoktor: It shouldn't be a new issue, it's just issue that wasn't caught by previous version (probably bugfix on pylint side). But yeah, I could have improved the commit message but as it was just a style fix.

PR #128, opened 2024-01-31
Drop the use of pipes

luckyh:
`pipes` was deprecated since version 3.11 so let's replace the use of that module with the alternatives.


luckyh:
cc @yanan-fu @lkotek

ldoktor (approved):
Thanks, seems to cover all places.

PR #130, opened 2024-05-14
Add grep_pipe

gbladislau:
This adds a new functionality to the ops_linux.py enabling greping on direct commands outputs.

### DIFF
--- a/aexpect/ops_linux.py
+++ b/aexpect/ops_linux.py
@@ -432,6 +432,29 @@ def grep(session, expr, path, check=False, flags="-a"):
     return output
 
 
+def grep_pipe(session, cmd, expr, check=False, flags="-a"):
+    """
+    Invoke ``grep`` remotely searching for an expression in a command output.
+
+    :param session: session to run the command on
+    :type session: ShellSession
+    :param str cmd: command whose output will be grepped
+    :param str expr: search expression
+    :param bool check: whether to quietly run grep for a boolean check
+    :param str flags: extra flags passed to ``grep`` on the command line
+    :returns: whether there is a match or not or what ``grep`` emits on stdout
+              if the check mode is disabled
+    :rtype: bool or str
+    :raises: ShellCmdError if the check mode is disabled and status is nonzero
+    """
+    grep_command = f"{cmd} | grep {flags} '{expr}'"
+    status, output = session.cmd_status_output(grep_command)
+    if check:
+        return status == 0
+    _, output = _process_status_output(grep_command, status, output)
+    return output
+
+
 ###############################################################################
 # utilities
 ###############################################################################


aexpect/ops_linux.py
```diff
@@ -432,6 +432,29 @@ def grep(session, expr, path, check=False, flags="-a"):
     return output
 
 
+def grep_pipe(session, cmd, pattern, flags = "", check = False):
+    """
+    Invoke ``grep`` remotely searching for an expression in a command output.
+
+    :param session: vm session
```
ldoktor, line 439: `:param session: session to run the command on` would be more consistent

aexpect/ops_linux.py
```diff
@@ -432,6 +432,29 @@ def grep(session, expr, path, check=False, flags="-a"):
...
+    :type session: ShellSession
+    :param str cmd: command that its output will be grepped
```
ldoktor, line 441: I'm not native but this sounds strange. How about `command whose output will be grepped`?

aexpect/ops_linux.py
```diff
@@ -432,6 +432,29 @@ def grep(session, expr, path, check=False, flags="-a"):
...
+    :param str pattern: pattern to grep
```
ldoktor, line 442: Could you please unify this with the `grep` param above? (`expr` rather than `pattern`)

aexpect/ops_linux.py
```diff
@@ -432,6 +432,29 @@ def grep(session, expr, path, check=False, flags="-a"):
...
+    :param str flags: extra flags passed to ``grep`` on the command line
+    :param bool check: whether to quietly run grep for a boolean check
```
ldoktor, line 444: Again, it'd be nice to unify this with the existing `grep` by changing the position of `check` and `flags`.

Something like `grep_pipe(session, cmd, expr, check, flags)` would be IMO better (or even `session, expr, cmd, check, flags` but logically the `cmd` should be first here. Not a strong preference over those two, though.)

ldoktor (commented):
Hello @gbladislau, glad to see a new face in aexpect. Is this a theoretical improvement or are you using this somewhere? I'm not sure how important this is since one can simply use `session.cmd_status("cmd | grep ...")` manually but if there is a project using this frequently, we can add this as long as it's consistent with the existing `grep` command (arguments order and docstrings).

gbladislau:
Hi @ldoktor I just saw your review. I use the avocado project on a daily basis at the company I work for. This is just a addition to the project since I used `session.cmd_status("cmd| grep ...")` a lot I think it would be good to have a grep for commands. I'll make the changes from your review and push them.

aexpect/ops_linux.py
gbladislau: Much better, I'm not native either you can see that.

ldoktor (commented):
Thanks, this looks good. Could you please squash it into a single commit so we can get this in?

gbladislau:
Done!

ldoktor (approved):
thanks, lgtm

PR #136, opened 2024-10-10
Provide Pyro5 compatibility to the remote door utility

pevogam:
The main adaptation that was needed comes down to dropping the previous expose-free approach and dynamically exposing each relevant function and class in addition to the previous autoproxy steps.

To allow for modules and classes that the shared remote object has not explicitly imported or are not detectable as needing exposing, the previous whitelist argument has been extended to allow exposing predefined classes and serializing predefined exceptions that could be used and raised during remote use.

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -838,7 +838,7 @@ def import_remote_exceptions(exceptions=None, modules=None):
     :type exceptions: [str] or None
     :param modules: full module paths whose custom exceptions will first be
                     detected and then automatically imported (optional)
-    :type exceptions: [str] or None
+    :type modules: [str] or None
 
     The deserialization by our Pyro backend requires the full module paths to
     each exception or module in order to correctly detect the exception type.
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -54,7 +54,7 @@ INTERFACE
 
 # disable import issues from optional dependencies or remote extra imports using
 # pylint: disable=E0401,C0415,W0212
-# disable too-many-* as we need them pylint: disable=R0912,R0913,R0914,R0915,C0302
+# disable too-many/few-* as we need them pylint: disable=R0903,R0912,R0913,R0914,R0915,R0917,C0302
 # ..todo:: we could reduce the disabled issues after more significant refactoring
 
 import os
@@ -70,8 +70,29 @@ import time
 # related to connectivity and perform further development on this utility.
 # os.environ["PYRO_LOGLEVEL"] = "DEBUG"
 try:
-    # noinspection PyPackageRequirements,PyUnresolvedReferences
-    import Pyro4
+    try:
+        # noinspection PyPackageRequirements,PyUnresolvedReferences
+        from Pyro5.compatibility import Pyro4
+        # noinspection PyPackageRequirements
+        from Pyro5 import server
+        # noinspection PyPackageRequirements
+        from Pyro5 import nameserver
+    except ImportError:
+        # noinspection PyPackageRequirements,PyUnresolvedReferences
+        import Pyro4
+        # Pyro5 compatibility does not support submodules so we need separate handling
+
+        class Server:
+            """Dummy class to replace a Pyro5 server."""
+            def __init__(self):
+                self.__dict__ = {}
+        server = Server()
+        server.expose = Pyro4.expose  # pylint: disable=W0201
+        server.is_private_attribute = Pyro4.util.is_private_attribute  # pylint: disable=W0201
+        # noinspection PyPackageRequirements
+        from Pyro4 import naming as nameserver
+        nameserver.start_ns = nameserver.startNS
+
 except ImportError:
     logging.warning("Remote object backend (Pyro4) not found, some functionality"
                     " of the remote door will not be available")
@@ -411,11 +432,13 @@ def set_subcontrol_parameter_dict(subcontrol, dict_name, value):
 
 
 @set_subcontrol
-def set_subcontrol_parameter_object(subcontrol, value):
+def set_subcontrol_parameter_object(subcontrol, object_type, value):
     """
     Prepare a URI to remote params for the remote control file.
 
     :param str subcontrol: path to the original control
+    :param str object_type: type of the object described as module path text,
+               for example "collections.UserDict"
     :param value: control parameters value
     :type value: Params object
     :returns: path to the modified control
@@ -438,7 +461,7 @@ def set_subcontrol_parameter_object(subcontrol, value):
 
     LOG.info("Sharing the test parameters over the network")
     Pyro4.config.AUTOPROXY = False
-    Pyro4.config.REQUIRE_EXPOSE = False
+    expose_remote_classes([object_type])
     try:
         pyro_daemon = Pyro4.Daemon(host=host_ip, port=1437)
         LOG.debug("Pyro4 daemon started successfully")
@@ -496,7 +519,8 @@ class DaemonLoop(threading.Thread):
         self.pyro_daemon.requestLoop()
 
 
-def get_remote_object(object_name, session=None, host="localhost", port=9090):
+def get_remote_object(object_name, session=None, host="localhost", port=9090,
+                      object_wl=None, expose_wl=None, serialize_wl=None):
     """
     Get a data object (visual or other) executing remotely or
     share one if none is available, generating control files along
@@ -508,6 +532,16 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
     :type session: RemoteSession or None
     :param str host: ip address of the local sharing server
     :param int port: port of the local name server
+    :param object_wl: whitelist of allowed functions/classes of the local (typically
+                      module) object as a tuple pair (functions, classes);
+                      expose and autoproxy all (do not filter) if None or empty
+    :type object_wl: ([str],[str]) or None
+    :param expose_wl: extra classes to expose as a pair of full module class paths
+                      and general module paths to expose all classes from
+    :type expose_wl: ([str],[str]) or None
+    :param serialize_wl: exceptions to serialize as a pair of full module class paths
+                         and general module paths to serialize all exceptions from
+    :type serialize_wl: ([str],[str]) or None
     :returns: proxy version of the remote object
     :rtype: Pyro4.Proxy
 
@@ -531,6 +565,10 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
     This method does not rely on any static (template) controls in order to
     work because the remote door takes care to reach back the local one.
     """
+    serialize_wl = serialize_wl or ([], [])
+    # additional exceptions to serialize
+    import_remote_exceptions(*serialize_wl)
+
     # noinspection PyUnresolvedReferences
     try:
         remote_object = Pyro4.Proxy(f"PYRONAME:{object_name}@{host}:{port}")
@@ -543,7 +581,10 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
         # if there is no door on the other side, open one
         _copy_control(session, os.path.abspath(__file__), is_utility=True)
         run_remote_util(session, "remote_door", "share_local_object",
-                        object_name, host=host, port=port, detach=True)
+                        object_name, host=host, port=port,
+                        object_wl=object_wl,
+                        expose_wl=expose_wl,
+                        detach=True)
         output, attempts = "", 10
         for _ in range(attempts):
             output = session.get_output()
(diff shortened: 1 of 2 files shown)


pevogam:
There are many modules out there that have `Too many positional arguments` and I believe it should be just a recommendation. I cannot see the isolation tests here in further details but they all pass locally.

aexpect/remote_door.py
```diff
@@ -508,6 +526,16 @@ def get_remote_object(object_name, session=None, host="localhost", port=9090):
...
     :param str host: ip address of the local sharing server
     :param int port: port of the local name server
+    :param object_wl: whitelist of allowed functions/classes of the local (typically
+                      module) object as a tuple pair (functions, classes);
+                      expose and autoproxy all (do not filter) if None or empty
+    :type object_wl: ([str],[str]) or None
+    :param expose_wl: extra classes to expose as a pair of full module class paths
+                      and general module paths to expose all classes from
+    :type expose_wl: ([str],[str]) or None
+    :param serialize_wl: exceptions to serialize as as a pair of full module class paths
```
ldoktor, line 536: typo: double `as as`

aexpect/remote_door.py
```diff
@@ -611,26 +643,35 @@ def get_remote_objects(session=None, host="localhost", port=0):
...
     :param port: port of the local sharing server
     :type port: int or str
+    :param object_wl: whitelist of allowed functions/classes of the local (typically
+                      module) object as a tuple pair (functions, classes);
+                      expose and autoproxy all (do not filter) if None or empty
+    :type object_wl: ([str],[str]) or None
+    :param expose_wl: extra classes to expose as a pair of full module class paths
+                      and general module paths to expose all classes from
+    :type expose_wl: ([str],[str]) or None
+    :param serialize_wl: exceptions to serialize as as a pair of full module class paths
```
ldoktor, line 662: typo: `as as`

aexpect/remote_door.py
```diff
@@ -696,12 +735,31 @@ class ModuleObject:  # pylint: disable=R0903
         """Module wrapped for transferability."""
 
     for fname, fobj in inspect.getmembers(module, inspect.isfunction):
-        if not whitelist or (object_name, fname) in whitelist:
+        if not object_wl[0] or fname in object_wl[0]:
```
ldoktor, line 738: IIUC the `whitelist` used to be `[(module1, function1), (module2, function2)]` and here you're changing it to `[(module1, module2), (function1, function2)]`?

aexpect/remote_door.py
```diff
@@ -838,7 +896,7 @@ def import_remote_exceptions(exceptions=None, modules=None):
     :type exceptions: [str] or None
     :param modules: full module paths whose custom exceptions will first be
                     detected and then automatically imported (optional)
-    :type exceptions: [str] or None
+    :type modules: [str] or None
```
ldoktor, line 899: good catch, ideally should be separate commit (but unless there are more we can close one eye and let it here ;-) )

aexpect/remote_door.py
```diff
@@ -438,7 +455,7 @@ def set_subcontrol_parameter_object(subcontrol, value):
 
     LOG.info("Sharing the test parameters over the network")
     Pyro4.config.AUTOPROXY = False
-    Pyro4.config.REQUIRE_EXPOSE = False
+    expose_remote_classes(["virttest.utils_params.Params"])
```
ldoktor, line 458: I don't think this belongs here. The aexpect project is generic, not tied to avocado-vt/virttest.

aexpect/remote_door.py
```diff
@@ -885,3 +943,52 @@ def recreate_exception(class_name, class_dict):
...
+            if base_cls in (object, int, float, bool, str, tuple, frozenset):
+                # known immutable classes should be skipped
+                continue
+            if inspect.ismethoddescriptor(base_cls):
+                # method descriptors should be skipped
+                continue
+            try:
+                server.expose(base_cls)
+            except (TypeError, AttributeError) as error:
+                logging.warning("Additional class exposing error: %s", error)
```
ldoktor, line 993: IIUC it's not `additional` but `parent` class, right?

tests/test_remote_door.py
```diff
@@ -153,6 +154,8 @@ def test_get_remote_object(self):
         self.session = mock.MagicMock(name='session')
         self.session.client = "ssh"
         remote_door.Pyro4 = mock.MagicMock()
+        remote_door.server = mock.MagicMock()
+        remote_door.nameserver = mock.MagicMock()
```
ldoktor, line 158: Since those are suppose to be created on import I think this setup should be cleared between tests. How about moving it to `setUp/tearDown` to ensure it's always as expected? (running a subset of tests or in different order might result in `remote_door` being different)

ldoktor (changes requested):
Hello @pevogam, thanks for keeping the remote door up and running. My only hard requirement is the `virttest.utils_params` removal as this is supposed to be generic project. I hope you'll find a better place for this tweak inside `avocado-vt` (could be a post-init step or whatever...).

As for the rest, I'd prefer having it split into multiple incremental changes, but since this is "an extension" and not core aexpect, it's not a strong requirement. It'd just be easier to follow as, if I got it right, some changes are not really related to the pyro change but are rather improvements or refinements.

aexpect/remote_door.py
pevogam: Rather to `([function1, function2], [module1, module2])` as documented on line 658.

pevogam: As the user typically access a remote object and this internally calls the sharing of a local object, there are few calls that have previously depended on this argument and the current change is putting it in actual use.

aexpect/remote_door.py
pevogam: I don't think it relates to a separate commit since the changes of exposing extra classes are very similar in nature to the changes of serializing extra exceptions as can be seen from the function signatures as well as the corresponding tests.

aexpect/remote_door.py
pevogam: Not quite, the mro will also include the starting class, here the `cls` itself. So what we will add is the extra class *and* all its base methods to provide a proper "network inheritance".

tests/test_remote_door.py
pevogam: The problem is that the imports might not always be available, after all isolation tests can be run in complete isolation. This then means that there is no easy way to just patch these and recover the original import (which might not exist) afterwards. But I will look for some way as you have a point here.

pevogam:
Hi @ldoktor and thanks so much for spending the effort to review this, I know such resources are precious. 

> Hello @pevogam, thanks for keeping the remote door up and running. My only hard requirement is the `virttest.utils_params` removal as this is supposed to be generic project. I hope you'll find a better place for this tweak inside `avocado-vt` (could be a post-init step or whatever...).

I have pushed changes regarding the extra hard-coding point and typos you mentioned.

> As for the rest, I'd prefer having it split into multiple incremental changes, but since this is "an extension" and not core aexpect, it's not a strong requirement. It'd just be easier to follow as, if I got it right, some changes are not really related to the pyro change but are rather improvements or refinements.

The refinements are all touching upon the same changes and we cannot make the remote door compatible with Pyro5 without them (or the tests and usage would break at any further intermediary steps) as the only significant refining I can think of is extending the local object sharing with additional exposing flexibility, all of which is a requirement for using Pyro5 in any way.

Regarding the CI, should we disable the `Too many positional arguments` globally since so many modules have this problem?

pevogam:
I will run some more integration tests during the weekend, perhaps I will separate the serialization part is the second major part that wasn't needed for Pyro5 - you are right.

aexpect/remote_door.py
ldoktor: What I meant is that this is a typo-fix and not actual change related to this commit.

aexpect/remote_door.py
ldoktor: Sure, I meant the same but wasn't explicit. My point is that `additional` seems like some extra added classes while here (IIUC) it's more about all classes that this class uses (so recursive parents). If so I think the message should be `Parent(s) class exponsing error` (or `Upper`?) rather than `Additional`. Again, take this just for consideration, no need to change it if you think `Additional` is more suitable. I'll definitely trust your opinion here since you wrote all of it.

aexpect/remote_door.py
```diff
@@ -70,8 +70,27 @@
...
+        from Pyro5.compatibility import Pyro4
+        # noinspection PyPackageRequirements
+        from Pyro5 import server
+        # noinspection PyPackageRequirements
+        from Pyro5 import nameserver
+    except ImportError:
+        # noinspection PyPackageRequirements,PyUnresolvedReferences
+        import Pyro4
+        # Pyro5 compatibility does not support submodules so we need separate handling
+        class Server:
```
ldoktor, line 84: Thanks, I like this version better than using mock in "production" code.

aexpect/remote_door.py
```diff
@@ -411,7 +430,7 @@ def set_subcontrol_parameter_dict(subcontrol, dict_name, value):
 
 
 @set_subcontrol
-def set_subcontrol_parameter_object(subcontrol, value):
+def set_subcontrol_parameter_object(subcontrol, object_type, value):
```
ldoktor, line 433: Would you please also add the `object_type` to the docstrings?

ldoktor (approved):
Thanks, this version looks better as for me it's mostly ready to be merged once it passes the CI (I haven't tested it functionally, just reviewed it). As for the too-many pylint failures, simple add them to the  `pylint: disable` on line 56. (still if you prefer to split it into multiple commits, feel free to do so, but as it's not core I don't object if you keep it as is...)

pevogam:
Thanks a lot @ldoktor, I will take a look for a second and hopefully final round of updates (from your feedback and some of our own integration testing) by some time tomorrow and then all of this will be finalized for you.

aexpect/remote_door.py
pevogam: No worries! Resolving this

aexpect/remote_door.py
pevogam: Ok, done and separated so I will resolve this.

aexpect/remote_door.py
pevogam: The `Parent` class exposing would be incorrect since we are exposing the initially given classes together with their parents. "Additional" here is in the sense that we don't just expose the class (or emulated module class) of the initial object but additional classes that it makes use of for which we will have to be explicit now. Thankfully these will be fewer cases with type hint annotated code overtaking more software projects in the future. The reason for this is that the above extra classes are also needed in addition to any classes that are either *defined or imported* in the original namespace, all of which will already be handled without this extra argument. Would you like me to document any of this further in the docstrings?

aexpect/remote_door.py
pevogam: It was necessitated by the unit test refactoring for the mock imports too. Resolving this then!

ldoktor:
The R0917 is related to the new (well 3 months old) pylint update so let's disable it here https://github.com/avocado-framework/aexpect/pull/137

As for the `noqa` this tag works for lgtm and CodeQL but pylint requires to explicitly list the errors by `pylint: disable=XXX`. If you do it on a separate line it'll disable it for the whole file (or until you re-enable it) but it's better to just put it on the line where you intend it to work (then it'll be only disabled for that line)

aexpect/remote_door.py
ldoktor: Nop, I was just curious about the error message but it seems you had a good reason for this name :-)

pevogam:
> The R0917 is related to the new (well 3 months old) pylint update so let's disable it here #137

Looks good! So it was a catch overall.

> As for the `noqa` this tag works for lgtm and CodeQL but pylint requires to explicitly list the errors by `pylint: disable=XXX`. If you do it on a separate line it'll disable it for the whole file (or until you re-enable it) but it's better to just put it on the line where you intend it to work (then it'll be only disabled for that line)

The problem is that `# pylint: disable=E301` which is what I originally tried ends up in
```
aexpect/remote_door.py:84:0: W0012: Unknown option value for 'disable', expected a valid pylint message and got 'E301' (unknown-option-value)
```
locally and in the ignore comment being, well, ignored by pylint in the CI. So I assumed that the custom linter here internally invokes pycodestyle.

pevogam:
> > > As for the `noqa` this tag works for lgtm and CodeQL but pylint requires to explicitly list the errors by `pylint: disable=XXX`. If you do it on a separate line it'll disable it for the whole file (or until you re-enable it) but it's better to just put it on the line where you intend it to work (then it'll be only disabled for that line)
> > 
> > 
> > The problem is that `# pylint: disable=E301` which is what I originally tried ends up in
> > ```
> > aexpect/remote_door.py:84:0: W0012: Unknown option value for 'disable', expected a valid pylint message and got 'E301' (unknown-option-value)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > locally and in the ignore comment being, well, ignored by pylint in the CI. So I assumed that the custom linter here internally invokes pycodestyle.
> 
> It's likely you have an older pylint, github uses 3.3.1. You can upgrade it by `pip install -U pylint --user`.

Maybe I miss something from what you originally meant but you can see that even the error ID is not compatible with the numbering done my pylint. The error also happens despite pylint giving its 10/10 grade. So it is strictly PEP8 related. However, I give up trying to convince the inspektor or frontend that `noqa` should have worked there and simply added the extra empty line. It might be something to bring up to the inspektor project but best to move on here -> CI is passing now.

pevogam:
I fixed another typo with an extra push. Considering aexpect has last been released more than a year ago what are your thoughts on another tag+release? Perhaps @lmr is the person to turn to regarding feedback about this?

ldoktor:
Thanks, looks good to me, let me ping Lucas WRT release.

PR #137, opened 2024-10-15
Makefile: Skip newly added R0917

ldoktor:
disable too-many-positional-arguments (R0917) in our CI as we don't intend to refactor the code to use named arguments instead.


pevogam (approved):
Very good addition to our general approach to disable any "too many" types of pylint errors. LGTM

Issue #127, opened 2024-01-26
Release aexpect 1.7.1


pevogam:
Since we have done some separation of functionality with Avocado VT and it has been tagged and released as 103.0 perhaps it would be a good idea to also release axpect now for easier traceability  and to provide any of collected changes to the users.

@ldoktor @lmr What do you think about this suggestion?

vfazio:
This would be useful for us since the usage of `pipes` was dropped and thus implicit support of Python 3.13

ldoktor:
Makes sense to me although I'd like to see https://github.com/avocado-framework/aexpect/pull/135 and https://github.com/avocado-framework/aexpect/pull/141 and https://github.com/avocado-framework/aexpect/pull/142 in the release, what do you think?

PR #141, opened 2025-01-03
Removing the escape sequence from the console outputs!

Anushree-Mathur:
Testcases fail as the output is having escape sequence in it which is not expected to be there.Removing this escape sequence fixes almost 2000+ testcases.
Signed-off-by: Anushree Mathur <anushree.mathur@linux.vnet.ibm.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -996,7 +996,9 @@ class Expect(Tail):
         def _get_last_nonempty_line(cont):
             nonempty_lines = [_ for _ in cont.splitlines() if _.strip()]
             if nonempty_lines:
-                return nonempty_lines[-1]
+                # Removing escape sequence from the last element of nonempty_lines
+                nonempty_last_no_escape = astring.strip_console_codes(nonempty_lines[-1])
+                return nonempty_last_no_escape
             return ""
 
         return self.read_until_output_matches(patterns,


Anushree-Mathur:
efore applying the patch i tried to print the outputs in between and found out the escape sequence issue.Many testcases were getting failed, 1 of the example is as follows:
```
[stdout] NEW NONEMPTY LIST []
[stdout]
[stdout] CONT in the read_until_last_line_matches
[stdout] Last login: Fri Jan  3 07:46:25 from 192.168.122.1
[stdout] ^[[?2004h[root@localhost ~]#
[stdout]
[stdout] nonempty lines in until last line matches ['Last login: Fri Jan  3 07:46:25 from 192.168.122.1', '\x1b[?2004h[root@localhost ~]# ']
[stdout]
[stdout] Nonempty lines are true in read_until_last ['Last login: Fri Jan  3 07:46:25 from 192.168.122.1', '\x1b[?2004h[root@localhost ~]# ']
[stdout] 
[stdout] \x1b[?2004h[root@localhost ~]#
```
```
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: STARTED
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: PASS (244.46 s)
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: STARTED
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: ERROR: Login timeout expired    (output: 'exceeded 240 s timeout') (672.89 s)
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: STARTED
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: PASS (5.10 s)
```


After applying the patch, testcases are getting passed:

```
[stdout] Nonempty lines are true in read_until_last ['Last login: Fri Jan  3 09:04:36 from 192.168.122.1', '\x1b[?2004h[root@localhost ~]# ']
[stdout]
[stdout] [root@localhost ~]#
```
Last element of the list gets used after removing the escape sequence. This is necessary to check for the the proper login.



```
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: STARTED
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: PASS (230.22 s)
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: STARTED
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: PASS (626.06 s)
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: STARTED
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: PASS (5.01 s)
```

ldoktor (commented):
Hello @Anushree-Mathur, this change makes sense to me as in `get_last_non_empty_line` we do expect console output which might be corrupted by console codes. My only concern is whether we shouldn't strip the full output as the current version might result in `""` output (when last nonempty line contains only escape code, it'll be identified as non-empty and then stripped).

I think I need an extra pair of eyes on this as it might affect the behaviour @smitterl or @pevogam can you please take a look at this?

Anushree-Mathur:
Hello @ldoktor, thanks alot for taking time and reviewing this commit! The much I have understood, IMO the output will surely be something with the escape sequence but if it is only containing escape sequence and stripping this comes as empty ouput, maybe that was expected? The output that is coming in the console and the way it is getting read, those are different! So it will take the output only until the last nonempty output that is correct according to the console but the escape code gets added in a way it gets read. Please correct me if i am wrong. 

Thanks in advance!

ldoktor:
> Hello @ldoktor, thanks alot for taking time and reviewing this commit! The much I have understood, IMO the output will surely be something with the escape sequence but if it is only containing escape sequence and stripping this comes as empty ouput, maybe that was expected? The output that is coming in the console and the way it is getting read, those are different! So it will take the output only until the last nonempty output that is correct according to the console but the escape code gets added in a way it gets read. Please correct me if i am wrong.
> 
> Thanks in advance!

I'm talking about situation like this `foo\nbar\n\x1b[33;1m`. With your modification it returns `""` while originally it returned `"\x1b[33;1m"`. I think it'd be better to strip the input so it becomes `foo\nbar\n` and only `"bar"` gets returned.

pevogam:
Hi, I will be able to test and review this towards the end of the week, I hope this is ok and thanks for the contribution!

Anushree-Mathur:
Thank you so much @pevogam @ldoktor for your time! I am getting your point now @ldoktor, I am also looking into this, will get back to you on this by tomorrow. Thank you for the clarification.

ldoktor:
I'm not sure I follow, can you please elaborate a bit more? The function you're modifying is inside a filter function `_get_last_nonempty_line` which is used by `read_until_last_line_matches` to only match the latest output from `Tail` process and is often used to read up to prompt (which should be the last non-empty line). Thank you for noticing that when the output contains extra characters (eg. when executing `echo -n whatever`) the last line might contain these  control characters (eg. `\x1b[?2004h`) which we can and probably should remove by `astring.strip_console_codes`.

My proposal was about another potential situation, where the console spits additional content after the last line which would again contain only control characters (eg. `Last login: Fri Jan  3 07:46:25 from 192.168.122.1\n[root@localhost ~]# \n\x1b[?2004h`). In such case the current code returns `"\x1b[?2004g"`, your code returns `""` and I'm proposing it should ignore those lines as well and return last non-empty one `"[root@localhost ~]# "`.

ldoktor:
Any updates on this @Anushree-Mathur?

ldoktor:
> @Anushree-Mathur Hi, May I ask how this problem is progressing? It seems the next aexpect release is waiting for this fix. @ldoktor

We have currently a few quite important pending PRs and then we'd like to make a release. So we don't really wait for this one, but it'd be nice to know whether you intend to get this into the next release or not.

xiaodwan:
@ldoktor thanks for your reply.  I'll wait for your plan.

Anushree-Mathur:
Hi @xiaodwan @ldoktor , sorry for the delay in reply! I was on leave. I will update the PR as soon as possible. Thank you for the patience.

pevogam:
Hi @Anushree-Mathur do you still have interest to update this pull request? I am asking since it is a blocker for #127.

pevogam:
> Hi @Anushree-Mathur do you still have interest to update this pull request? I am asking since it is a blocker for #127.

To update the status from my own last reply - this is no longer a blocker and the release was completed. The pull request is thus lower priority again.

PR #142, opened 2025-01-05
Removing escape sequence from output in cmd_output function!

Anushree-Mathur:
In console outputs we are getting escape sequences because of which most of the testcases are failing to use the output of few commands. This PR removes the escape sequence first then use the output.This PR fixes multiple testcases.
Signed-off-by: Anushree Mathur <anushree.mathur@linux.vnet.ibm.com>

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1188,7 +1188,7 @@ class ShellSession(Expect):
                                                  print_func)[1]
 
     def cmd_output(self, cmd, timeout=60, internal_timeout=None,
-                   print_func=None, safe=False):
+                   print_func=None, safe=False, strip_console_codes=False):
         """
         Send a command and return its output.
 
@@ -1203,7 +1203,10 @@ class ShellSession(Expect):
                 error messages that make read_up_to_prompt to timeout. Let's
                 try to be a little more robust and send a carriage return, to
                 see if we can get to the prompt when safe=True.
-
+        :param strip_console_codes: Whether to remove the escape sequence from the output.
+                In serial sessions, there are escape sequences present. If it is not
+                expected while reading the output remove it by passing
+                serial_console_codes = True.
         :return: The output of cmd
         :raise ShellTimeoutError: Raised if timeout expires
         :raise ShellProcessTerminatedError: Raised if the shell process
@@ -1211,7 +1214,7 @@ class ShellSession(Expect):
         :raise ShellError: Raised if an unknown error occurs
         """
         if safe:
-            return self.cmd_output_safe(cmd, timeout)
+            return self.cmd_output_safe(cmd, timeout, strip_console_codes)
         session_tag = f"[{self.output_prefix}] " if self.output_prefix else ""
         LOG.debug("%sSending command: %s", session_tag, cmd)
         self.read_nonblocking(0, timeout)
@@ -1229,10 +1232,13 @@ class ShellSession(Expect):
             raise ShellError(cmd, output) from error
 
         # Remove the echoed command and the final shell prompt
+        if strip_console_codes:
+            # Removing the escape sequence
+            out = astring.strip_console_codes(out)
         return self.remove_last_nonempty_line(self.remove_command_echo(out,
                                                                        cmd))
 
-    def cmd_output_safe(self, cmd, timeout=60):
+    def cmd_output_safe(self, cmd, timeout=60, strip_console_codes=False):
         """
         Send a command and return its output (serial sessions).
 
@@ -1244,7 +1250,10 @@ class ShellSession(Expect):
         :param cmd: Command to send (must not contain newline characters)
         :param timeout: The duration (in seconds) to wait for the prompt to
                 return
-
+        :param strip_console_codes: Whether to remove the escape sequence from the output.
+                In serial sessions, there are escape sequences present. If it is not
+                expected while reading the output remove it by passing
+                serial_console_codes = True.
         :return: The output of cmd
         :raise ShellTimeoutError: Raised if timeout expires
         :raise ShellProcessTerminatedError: Raised if the shell process
@@ -1278,6 +1287,9 @@ class ShellSession(Expect):
             raise ShellTimeoutError(cmd, out)
 
         # Remove the echoed command and the final shell prompt
+        if strip_console_codes:
+            # Removing the escape sequence
+            out = astring.strip_console_codes(out)
         return self.remove_last_nonempty_line(self.remove_command_echo(out,
                                                                        cmd))
 


Anushree-Mathur:
In total 2000+ testcases are failing but to give an example I am mentioning the following testcase!
Before applying the patch:
```
[stdout] escape output cat /proc/uptime
[stdout] ^[[?2004l168.18 720.90
[stdout] ^[[?2004h[root@localhost ~]#
```

```
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: STARTED
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: PASS (232.38 s)
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: STARTED
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: ERROR: could not convert string to float: '\x1b[?2004l168.18' (443.02 s)
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: STARTED
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: PASS (5.08 s)
```

After applying the patch:
```
[stdout] output cat /proc/uptime
[stdout] 168.18 720.90
[stdout] [root@localhost ~]# 
```

```
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: STARTED
 (1/3) io-github-autotest-qemu.unattended_install.import.import.default_install.aio_native: PASS (236.23 s)
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: STARTED
 (2/3) type_specific.io-github-autotest-libvirt.virsh.migrate.there_and_back_with_numa.with_numa_only.with_mem_hotplug.with_cpu_hotplug.with_hotplug.hotplug_after_migration.compat_migration.non_compat_migration.with_postcopy: PASS (614.98 s)
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: STARTED
 (3/3) io-github-autotest-libvirt.remove_guest.without_disk: PASS (5.07 s)
```

ldoktor (commented):
I don't think the `strip_console_codes` should be used at all times as `get_output` might be used to interact with any kind of output. IMO the consumers should decide whether they need to strip the console codes or leave them present in the output.

It's true that people interacting with odd outputs might use `read_nonblocking` directly to get the full output, but having `get_output` seems convenient and some only need to check the output after the execution (eg. tests needing to check whether arrows were used correctly).

I mean if there were thousands of lines where people need to clean the output, I'd consider adding an argument to do so, but doing it out of the box seems like removing useful feature for lower-level usage.

@smitterl, @pevogam would you have any opinions?

Anushree-Mathur:
Hello @ldoktor , thanks alot for taking time and reviewing this commit! I understand your point, so do you suggest that we do it locally in the functions where these outputs are being used? Could you please guide me how i can proceed otherwise?

Thanks in advance!

ldoktor:
> Hello @ldoktor , thanks alot for taking time and reviewing this commit! I understand your point, so do you suggest that we do it locally in the functions where these outputs are being used? Could you please guide me how i can proceed otherwise?
> 
> Thanks in advance!

What I meant is to do it explicitly in the test itself:

```python
from aexpect.utils.astring import strip_console_codes

...

out = session.cmd_output(...)
out_safe = strip_console_codes(out)
```

But if it proves to be massively used we might consider support for `session.cmd_output(..., strip_console_codes=True)` which would allow doing this inside the cmd_output.

Anushree-Mathur:
Hello @ldoktor , understood your point clearly! Thank you so much. Will make the changes in the testcase itself.

pevogam:
Hi, I will be able to test and review this towards the end of the week, I hope this is ok and thanks for the contribution!

Anushree-Mathur:
Hi @ldoktor @pevogam , I have tried modifying my PR in the following way as suggested by @ldoktor! Many places cmd_output is used so I changed the code here and tried to run  the same tescase after applying the strip_console_code=True in the local code where I wanted the O/P to be replaced and it worked totally fine! I would really like to have some suggestions here if I can proceed in this manner! Thanks in advance for your time.
```
    def cmd_output(self, cmd, timeout=60, internal_timeout=None,
                   print_func=None, safe=False, strip_console_codes=False):
        """
        Send a command and return its output.

        :param cmd: Command to send (must not contain newline characters)
        :param timeout: The duration (in seconds) to wait for the prompt to
                return
        :param internal_timeout: The timeout to pass to read_nonblocking
        :param print_func: A function to be used to print the data being read
                (should take a string parameter)
        :param safe: Whether using safe mode when execute cmd.
                In serial sessions, frequently the kernel might print debug or
                error messages that make read_up_to_prompt to timeout. Let's
                try to be a little more robust and send a carriage return, to
                see if we can get to the prompt when safe=True.

        :return: The output of cmd
        :raise ShellTimeoutError: Raised if timeout expires
        :raise ShellProcessTerminatedError: Raised if the shell process
                terminates while waiting for output
        :raise ShellError: Raised if an unknown error occurs
        """
        if safe:
            return self.cmd_output_safe(cmd, timeout)
        session_tag = f"[{self.output_prefix}] " if self.output_prefix else ""
        LOG.debug("%sSending command: %s", session_tag, cmd)
        self.read_nonblocking(0, timeout)
        self.sendline(cmd)
        try:
            out = self.read_up_to_prompt(timeout, internal_timeout, print_func)
        except ExpectTimeoutError as error:
            output = self.remove_command_echo(error.output, cmd)
            raise ShellTimeoutError(cmd, output) from error
        except ExpectProcessTerminatedError as error:
            output = self.remove_command_echo(error.output, cmd)
            raise ShellProcessTerminatedError(cmd, error.status, output) from error
        except ExpectError as error:
            output = self.remove_command_echo(error.output, cmd)
            raise ShellError(cmd, output) from error
        # Remove the echoed command and the final shell prompt
        if strip_console_codes:
            return self.remove_last_nonempty_line(self.remove_command_echo(astring.strip_console_codes(out),
                                                                       cmd))
        else:
            return self.remove_last_nonempty_line(self.remove_command_echo(out,
                                                                       cmd))
 ```

**Comment by pevogam** [comment]
Hi @Anushree-Mathur, could you perhaps provide a diff instead so that it is easy to parse what modification entails?

**Comment by Anushree-Mathur** [comment]
Hi @pevogam , here is the git diff result:
```
index e0460f4..6fa9f1e 100644
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1188,7 +1188,7 @@ class ShellSession(Expect):
                                                  print_func)[1]
 
     def cmd_output(self, cmd, timeout=60, internal_timeout=None,
-                   print_func=None, safe=False):
+                   print_func=None, safe=False, strip_console_codes=False):
         """
         Send a command and return its output.
 
@@ -1227,9 +1227,12 @@ class ShellSession(Expect):
         except ExpectError as error:
             output = self.remove_command_echo(error.output, cmd)
             raise ShellError(cmd, output) from error
-
         # Remove the echoed command and the final shell prompt
-        return self.remove_last_nonempty_line(self.remove_command_echo(out,
+        if strip_console_codes:
+            return self.remove_last_nonempty_line(self.remove_command_echo(astring.strip_console_codes(out),
+                                                                       cmd))
+        else:
+            return self.remove_last_nonempty_line(self.remove_command_echo(out,
                                                                        cmd))
 
     def cmd_output_safe(self, cmd, timeout=60):
```

**Comment by ldoktor** [comment]
This seems fine by me, provided the code is polished a bit. Mainly it needs to tackle the `cmd_output_safe` as well as the normal `cmd_output` and I'd prefer having just a single return statement with the `out` variable being modified before the return if asked for. @pevogam does this new argument sounds fine to you?

**Comment by pevogam** [comment]
> I don't think the `strip_console_codes` should be used at all times as `get_output` might be used to interact with any kind of output. IMO the consumers should decide whether they need to strip the console codes or leave them present in the output.
> 
> It's true that people interacting with odd outputs might use `read_nonblocking` directly to get the full output, but having `get_output` seems convenient and some only need to check the output after the execution (eg. tests needing to check whether arrows were used correctly).
> 
> I mean if there were thousands of lines where people need to clean the output, I'd consider adding an argument to do so, but doing it out of the box seems like removing useful feature for lower-level usage.

I second that opinion. This change is invasive and should preserve the behavior for anyone not wishing to remove the console codes and likely has their own handling logic down the line.

> Hi @pevogam , here is the git diff result:

I think this diff is much more reasonable and perhaps you could push the suggested change for easier review now?

> This seems fine by me, provided the code is polished a bit. Mainly it needs to tackle the `cmd_output_safe` as well as the normal `cmd_output` and I'd prefer having just a single return statement with the `out` variable being modified before the return if asked for. @pevogam does this new argument sounds fine to you?

I can definitely do in-line commenting and recommendations once I see this pushed but @Anushree-Mathur it would be great if you take in @ldoktor's recommendation before the push so we can save on the feedback cycles.

Also, I wonder what is the difference between this PR and #141? Could we instead close one of them and focus on just one pull request regarding escaping such characters?

**Comment by Anushree-Mathur** [comment]
Hi @pevogam @ldoktor , thank you so very much for the patience and your time in reviewing my PRs! I have understood most of the points made by both of you. I will surely push these changes by tomorrow. Few doubts I have, will clear it once I push these new changes.

**Comment by ldoktor** [comment]
Any updates on this @Anushree-Mathur?

**Comment by Anushree-Mathur** [comment]
Hi @ldoktor , apologies for being late! Got loaded with work. Will update this PR by tonight for sure! Thank you so much for the patience.

**Comment by ldoktor** [comment]
Sure, that is understandable. We are considering a release which is why I'm asking. No need to rush things, though...

**Comment by Anushree-Mathur** [comment]
Thank you so much for understanding @ldoktor !

File: aexpect/client.py [hunk 1]
```diff
@@ -259,7 +259,8 @@ def _get_aexpect_helper(self, helper_cmd, pass_fds, echo, command):
             full_output += output
             sub_status = sub.poll()
             if sub_status is not None:
-                raise ExpectProcessTerminatedError(pattern, sub_status, full_output)
+                raise ExpectProcessTerminatedError(
+                    pattern, sub_status, full_output)
```
pevogam, line 263: Was this change intentional? I think this is incompatible with black linting and doesn't relate to the topic of this pull request.

aexpect/client.py
```diff
@@ -1203,15 +1204,19 @@ def cmd_output(self, cmd, timeout=60, internal_timeout=None,
                 error messages that make read_up_to_prompt to timeout. Let's
                 try to be a little more robust and send a carriage return, to
                 see if we can get to the prompt when safe=True.
-
+        :param strip_console_codes: Whether remove the escape sequence from the output
```
pevogam, line 1207: Whether to remove the ... followed by a full stop.

aexpect/client.py
```diff
@@ -1203,15 +1204,19 @@ def cmd_output(self, cmd, timeout=60, internal_timeout=None,
...
+                In serial session there are escape sequences present,if it is not
```
pevogam, line 1208: In serial sessions.. with a space after the comma.

aexpect/client.py
```diff
@@ -1203,15 +1204,19 @@ def cmd_output(self, cmd, timeout=60, internal_timeout=None,
...
+                expected while reading the output remove it by passing
+                serial_console_codes = True
         :return: The output of cmd
         :raise ShellTimeoutError: Raised if timeout expires
         :raise ShellProcessTerminatedError: Raised if the shell process
                 terminates while waiting for output
         :raise ShellError: Raised if an unknown error occurs
+
```
pevogam, line 1216: This empty line is not needed

aexpect/client.py
```diff
@@ -1223,16 +1228,22 @@ def cmd_output(self, cmd, timeout=60, internal_timeout=None,
             raise ShellTimeoutError(cmd, output) from error
         except ExpectProcessTerminatedError as error:
             output = self.remove_command_echo(error.output, cmd)
-            raise ShellProcessTerminatedError(cmd, error.status, output) from error
+            raise ShellProcessTerminatedError(
+                cmd, error.status, output) from error
```
pevogam, line 1232: Similar statement to the "black" statement above

aexpect/client.py
```diff
@@ -1244,6 +1255,10 @@ def cmd_output_safe(self, cmd, timeout=60):
         :param cmd: Command to send (must not contain newline characters)
         :param timeout: The duration (in seconds) to wait for the prompt to
                 return
+        :param strip_console_codes: Whether remove the escape sequence from the output
+                In serial session there are escape sequences present,if it is not
```
pevogam, line 1259: Same adaptations would be needed here.

aexpect/client.py
```diff
@@ -1223,16 +1228,22 @@ def cmd_output(self, cmd, timeout=60, internal_timeout=None,
...
         except ExpectError as error:
             output = self.remove_command_echo(error.output, cmd)
             raise ShellError(cmd, output) from error
 
         # Remove the echoed command and the final shell prompt
+        if strip_console_codes:
+            # Removing the escape sequence
+            output_no_escape = astring.strip_console_codes(out)
+            return self.remove_last_nonempty_line(self.remove_command_echo(output_no_escape,
+                                                                           cmd))
```
pevogam, line 1242: How about a much simpler
```python
out = astring.strip_console_codes(out) if strip_console_codes else out
```
?

aexpect/client.py
```diff
@@ -1278,6 +1293,11 @@ def cmd_output_safe(self, cmd, timeout=60):
             raise ShellTimeoutError(cmd, out)
 
         # Remove the echoed command and the final shell prompt
+        if strip_console_codes:
+            # Removing the escape sequence
+            output_no_escape = astring.strip_console_codes(out)
+            return self.remove_last_nonempty_line(self.remove_command_echo(output_no_escape,
+                                                                           cmd))
```
pevogam, line 1300: Similar simplification to above

pevogam (commented):
Hi @Anushree-Mathur, this pull request definitely enhanced the session capabilities of aexpect, I left mostly a few typos to fix and a simplification suggestion.

aexpect/client.py
ldoktor: Well that's an extra assignment in case `strip_console_codes` is `False`. But I do like the idea of having just one return as they are mainly the same. How about:

```
if strip_console_codes:
    out = astring.strip_console_codes(out)
```

pevogam: It is a noop assignment and the above is a very pythonic way to perform such variable assignments but of course your suggestion is also fine.

Anushree-Mathur:
Hi @ldoktor @pevogam , thank you so much for the suggestions and your time, I have looked into this and will update this PR by afternoon!

aexpect/client.py
Anushree-Mathur: Oh I am not sure how did this happen @pevogam , I will remove this too.

ldoktor (commented):
Thanks, looks almost good. Could you please remove the empty line as suggested by Plamen so we're good to go?

pevogam (approved):
I also approve the rest of the changes just like @ldoktor here and have the same final comments.

Anushree-Mathur:
Hi @ldoktor @pevogam , thank you so much for your reviews and time. Removed the extra line also, lemme know if any further changes required! Thanks in advance.

PR #144, opened 2025-02-26
Add rsync client for remote copy operations

pevogam:
This commit adds support for a new rsync client for remote copy operations which is especially important now that the SCP protocol is being deprecated.

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -140,25 +140,29 @@ class SCPError(TransferError):
     """Remote error related to transfer using SCP."""
 
 
-class SCPAuthenticationError(SCPError):
+class RsyncError(TransferError):
+    """Remote error related to transfer using rsync."""
+
+
+class AuthenticationError(SCPError):
     """Remote error related to transfer authentication using SCP."""
 
 
-class SCPAuthenticationTimeoutError(SCPAuthenticationError):
+class AuthenticationTimeoutError(AuthenticationError):
     """Remote error related to transfer authentication timeout using SCP."""
 
     def __init__(self, output):
         super().__init__("Authentication timeout expired", output)
 
 
-class SCPTransferTimeoutError(SCPError):
+class TransferTimeoutError(SCPError):
     """Remote error related to transfer timeout using SCP."""
 
     def __init__(self, output):
         super().__init__("Transfer timeout expired", output)
 
 
-class SCPTransferFailedError(SCPError):
+class TransferFailedError(SCPError):
     """Remote error related to transfer failure using SCP."""
 
     def __init__(self, status, output):
@@ -381,7 +385,6 @@ def remote_login(client, host, port, username, password, prompt, linesep="\n",
     :raise LoginError: If using ipv6 linklocal but not assign an interface that
                        the neighbour attached
     :raise LoginBadClientError: If an unknown client is requested
-    :raise: Whatever handle_prompts() raises
     :return: A RemoteSession object.
     """
     extra_params = []
@@ -457,7 +460,6 @@ def wait_for_login(client, host, port, username, password, prompt,
     :interface: The interface the neighbours attach to
                 (only use when using ipv6 linklocal address.)
     :see: remote_login()
-    :raise: Whatever remote_login() raises
     :return: A RemoteSession object.
     """
     LOG.debug("Attempting to log into %s:%s using %s (timeout %ds)",
@@ -483,14 +485,14 @@ def wait_for_login(client, host, port, username, password, prompt,
                         user_known_hosts_file=user_known_hosts_file)
 
 
-def _remote_scp(
-        session, password_list, transfer_timeout=600, login_timeout=300):
+def _remote_copy(
+        session, password_list, transfer_timeout=600, login_timeout=300, method="scp"):
     """
-    Transfer files using SCP, given a command line.
+    Transfer files using SCP or rsync, given a command line.
 
-    Transfer file(s) to a remote host (guest) using SCP.  Wait for questions
-    and provide answers.  If login_timeout expires while waiting for output
-    from the child (e.g. a password prompt), fail.  If transfer_timeout expires
+    Transfer file(s) to a remote host (guest) using SCP or rsync. Wait for questions
+    and provide answers. If login_timeout expires while waiting for output
+    from the child (e.g. a password prompt), fail. If transfer_timeout expires
     while waiting for the transfer to complete, fail.
 
     :param session: An Expect or RemoteSession instance to operate on
@@ -500,9 +502,10 @@ def _remote_scp(
     :param login_timeout: The maximal time duration (in seconds) to wait for
             each step of the login procedure (i.e. the "Are you sure" prompt or
             the password prompt)
-    :raise SCPAuthenticationError: If authentication fails
-    :raise SCPTransferTimeoutError: If the transfer fails to complete in time
-    :raise SCPTransferFailedError: If the process terminates with a nonzero
+    :param method: The method to use for transfer ("scp" or "rsync")
+    :raise AuthenticationError: If authentication fails
+    :raise TransferTimeoutError: If the transfer fails to complete in time
+    :raise TransferFailedError: If the process terminates with a nonzero
             exit code
     :raise SCPError: If some other error occurs
     """
@@ -510,7 +513,7 @@ def _remote_scp(
     timeout = login_timeout
     authentication_done = False
 
-    scp_type = len(password_list)
+    transfer_type = len(password_list)
 
     while True:
         try:
@@ -527,9 +530,9 @@ def _remote_scp(
                     session.sendline(password_list[password_prompt_count])
                     password_prompt_count += 1
                     timeout = transfer_timeout
-                    if scp_type == 1:
+                    if transfer_type == 1:
                         authentication_done = True
-                elif password_prompt_count == 1 and scp_type == 2:
+                elif password_prompt_count == 1 and transfer_type == 2:
                     LOG.debug("Got password prompt, sending '%s'",
                               password_list[password_prompt_count])
                     session.sendline(password_list[password_prompt_count])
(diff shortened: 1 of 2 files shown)


aexpect/remote.py
```diff
@@ -500,17 +502,18 @@ def _remote_scp(
...
+    :raise TransferFailedError: If the process terminates with a nonzero
             exit code
     :raise SCPError: If some other error occurs
     """
     password_prompt_count = 0
     timeout = login_timeout
     authentication_done = False
 
-    scp_type = len(password_list)
+    transfer_type = len(password_list)
```
ldoktor, line 516: Well I wouldn't call it scp_type nor transfer_type, it's just number of passwords user provided. But this was here before so no obligation to change that (I was actually surprised by the whole function, could use some love, but I haven't seen anything obviously dangerous so again, no need to change at this point)

aexpect/remote.py
```diff
@@ -718,8 +720,142 @@ def scp_between_remotes(src, dst, port, s_passwd, d_passwd, s_name, d_name,
...
+    :param local_path: Path on the local machine where we are copying from
+    :param remote_path: Path on the remote machine where we are copying to
+    :param limit: Speed limit of file transfer.
+    :param log_filename: If specified, log all output to this file
+    :param log_function: If specified, log all output using this function
+    :param timeout: The time duration (in seconds) to wait for the transfer
+                    to complete.
+    :param interface: The interface the neighbours attach to (only use when using
+                      ipv6 linklocal address).
+    :param directory: True to copy recursively if the directory to rsync
```
ldoktor, line 746: this seems more important then `interface`, shouldn't it go first?

aexpect/remote.py
```diff
@@ -718,8 +720,142 @@ def scp_between_remotes(src, dst, port, s_passwd, d_passwd, s_name, d_name,
...
+    :param timeout: The time duration (in seconds) to wait for the transfer
+                    to complete.
+    :param interface: The interface the neighbours attach to (only use when
+                      using ipv6 linklocal address).
+    :param directory: True to copy recursively if the directory to rsync
+    :raise: Whatever remote_rsync() raises
+    """
+    if limit:
+        limit = f"--bwlimit={limit}"
+    if host and host.lower().startswith("fe80"):
```
ldoktor, line 793: This is repeated several-times, how about adding a function for that?

aexpect/remote.py
```diff
@@ -718,8 +720,142 @@ def scp_between_remotes(src, dst, port, s_passwd, d_passwd, s_name, d_name,
...
+        if not interface:
+            raise RsyncError("When using ipv6 linklocal address must assign, ",
+                             "the interface the neighbour attache")
+        host = f"{host}%{interface}"
+
+    command = "rsync"
+    if directory:
+        command = f"{command} -r"
```
ldoktor, line 801: how about `command += " -r"`?

aexpect/remote.py
```diff
@@ -718,8 +720,142 @@ def scp_between_remotes(src, dst, port, s_passwd, d_passwd, s_name, d_name,
...
+    command += (f" -avz -e 'ssh -p {port} -o UserKnownHostsFile=/dev/null "
+                f"-o StrictHostKeyChecking=no' {limit} "
+                f"{username}@{host}:{quote_path(remote_path)} {shlex.quote(local_path)}")
```
ldoktor, line 804: All of this is common to all three methods. How about splitting it into common part and then remote definition and share the common code that way?

I mean something like:

```python
def rsync_from_remote(...):
    command = _get_rsync_cmd(...)
    command += _get_rsync_host(...)
    command += "{shlex.quote(local_path)}"

def rsync_between_remotes(...):
    command = _get_rsync_cmd(...)
    command += _get_rsync_host(...)
    command += _get_rsync_host(...)
```

note it's not a strong request, I just think the less code one writes the easier is it to maintain it....

tests/test_remote.py
```diff
@@ -0,0 +1,122 @@
...
+            "ssh  -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -p 22"
+            " -o PreferredAuthentications=password user@127.0.0.1",
+        )
+
+    @mock.patch("aexpect.remote._remote_copy")
+    def test_scp_to_remote(self, mock_remote_copy):
+        remote.scp_to_remote(
+            "127.0.0.1", 22, "user", "pass", "/local/path", "/remote/path"
+        )
+        mock_remote_copy.assert_called_once_with(mock.ANY, ["pass"], 600, 300, "scp")
```
ldoktor, line 67: I think it'd make sense to check the `session.command` to record the expected behaviour, what do you think?

ldoktor (approved):
Hello Plamen, I provided some suggestions but there is no real blocker. The renamed exceptions might bring some difficulties to users, but I'm not sure how many rely on this so I leave it on your consideration. The same with the argument order as even the `scp` methods exhaust all the system arguments before the directory one so it is consistent as is (while not that user-friendly :D).

All in all, let me know what do you think about the exceptions and if you say it's fine I'm ready to merge it as is. But don't say nobody warned you then ;-)

aexpect/remote.py
```diff
@@ -567,19 +571,19 @@ def remote_scp(command, password_list, log_filename=None, log_function=None,
     :param login_timeout: The maximal time duration (in seconds) to wait for
             each step of the login procedure (i.e. the "Are you sure" prompt
             or the password prompt)
-    :raise: Whatever _remote_scp() raises
     """
-    LOG.debug("Trying to SCP with command '%s', timeout %ss",
+    LOG.debug("Trying to rsync or SCP with command '%s', timeout %ss",
```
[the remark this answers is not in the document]
pevogam: Sounds good

aexpect/remote.py
pevogam: We tend to avoid changing the order of parameters for users that don't rely on keywords and this is simply an adaptation of the `scp_from_remote` which has the same order (and likely user code assuming this).

aexpect/remote.py
pevogam: Sure

aexpect/remote.py
pevogam: I would say let's see if IPv6 really becomes properly supported in the future which would prompt some refactoring of such specialized validation.

aexpect/remote.py
```diff
@@ -140,25 +140,29 @@ class SCPError(TransferError):
     """Remote error related to transfer using SCP."""
 
 
-class SCPAuthenticationError(SCPError):
+class RsyncError(TransferError):
```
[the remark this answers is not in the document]
pevogam: I tried this first (separate rsync and scp hierarchies derived from the same base exceptions) but a lot of the refactored code which is nearly identical would have to be split into multiple versions for the only reason of raising a different exception type without much additional meaning (e.g. structure of the exception). So I ended up opting for using a simpler set of exceptions just like the "copy" instead of "SCP or rsync" in the above logging message.

tests/test_remote.py
pevogam: Sure, done

aexpect/remote.py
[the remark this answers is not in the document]
pevogam: So the point rather in the rafactoring and not as much in breaking changes. As you correctly observe above though, we could also break but improve the argument order regarding the `recursive` value, so let's do that too to at least make all of this somewhat nicer.

aexpect/remote.py
pevogam: In the future we might end up dropping scp entirely and just leaving rsync rendering such refactoring futile at present so I would keep the changes as additive as possible right now. Let's see after some more use of the remote module if it would make more sense to rather simplify it and maximize reusability.

pevogam:
Hi @ldoktor let's see how many people will rather thell me they warned me about changing the order of the directory argument which I did in a separate commit (as it is not directly related to the rsync extention). Other than that I hope I have addressed most if not all of your comments.

aexpect/remote.py
ldoktor: that seems reasonable

tests/test_remote.py
```diff
@@ -0,0 +1,146 @@
...
+            "dst_user",
+            "/src/path",
+            "/dst/path",
+        )
+        mock_remote_copy.assert_called_once_with(
+            mock.ANY, ["src_pass", "dst_pass"], 600, 300, "rsync"
+        )
+        self.assertEqual(
+            mock_remote_copy.call_args[0][0].command,
+            r"rsync -r -avz -e 'ssh -p 22 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'  src_user@src_host:/src/path dst_user@dst_host:/dst/path"
```
ldoktor, line 145:
```
>>> remote.rsync_between_remotes("127.0.0.1", "localhost", 22, "", "", "medic", "medic", "/tmp/a", "/tmp/b")
Traceback (most recent call last):
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/remote.py", line 520, in _remote_copy
    match, text = session.read_until_last_line_matches(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/client.py", line 1002, in read_until_last_line_matches
    return self.read_until_output_matches(patterns,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/client.py", line 940, in read_until_output_matches
    raise ExpectProcessTerminatedError(patterns, self.get_status(),
aexpect.exceptions.ExpectProcessTerminatedError: Process terminated while looking for patterns ['[Aa]re you sure', '[Pp]assword:\\s*$', 'lost connection']    (status: 1,    output: 'The source and destination cannot both be remote.\nrsync error: syntax or usage error (code 1) at main.c(1415) [Receiver=3.4.1]\n')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/remote.py", line 857, in rsync_between_remotes
    return remote_copy(command, password_list,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/remote.py", line 586, in remote_copy
    _remote_copy(session, password_list, transfer_timeout, login_timeout, method)
  File "/home/medic/Work/Projekty/avocado/aexpect/aexpect/remote.py", line 555, in _remote_copy
    raise TransferFailedError(error.status, error.output) from error
aexpect.remote.TransferFailedError: SCP transfer failed    (status: 1,    output: 'The source and destination cannot both be remote.\nrsync error: syntax or usage error (code 1) at main.c(1415) [Receiver=3.4.1]\n')
```

So it looks like you'll have to deal with this by either relying on `scp` or `ssh` to one of those and initiate the rsync from there (which might not cover the scenario where remotes don't have direct connection, then you'd have to tunel it through...)

pevogam: Ah interesting, in fact I have never used any copying between remotes and simply assumed rsync supports this as much as scp. If you want we can simply remove the support for this functionality then. The scp between remotes will still remain of course. What do you think?

ldoktor: makes sense to me. Better to be explicit than trying to hack something that might or might not work most of the time...

ldoktor (approved):
thanks, you are brave but prepare to suffer the consequences if anyone starts complaining about breakage. In main aexpect such changes would take longer but remote module is staging so let's get this in and benefit of cleaner api.

pevogam:
Ok let's see, I have also done quite some extensive integration testing and all our VT tests are passing with the rsync backend so at least this should be fine.

PR #146, opened 2025-04-03
Fix an fstring formatting for the remote linux ops

pevogam:


pevogam:
@ldoktor Just a minor fix here before the release. @lmr If you are not around could we provide access to @ldoktor or someone else for an annual release?

aexpect/ops_linux.py
```diff
@@ -397,7 +397,7 @@ def sed_replace(session, pattern, replacement, filename):
     Using sed is convenient, but it does not error out when nothing is replaced
     and this can lead to false-positives when used inside tests.
     """
-    LOG.info("Replacing {pattern} with {replacement} in {filename}")
+    LOG.info("Replacing %s with %s in %s", pattern, replacement, filename)
```
ldoktor, line 400: Well f-strings are recommended these days so how about adding the `f` before the string to make it work like a proper f-string?

pevogam: Yup, this is exactly what I did originally but aexpect has a linter that would fail if one uses fstrings with logging (not the recommended way according to some standards) and all other aexpect logging messages use this formatting (and the linter passes for them) so I settled for what is most compatible with the project code base.

ldoktor: Ah, I didn't noticed. Thanks, then this is definitely better (to log variables in case of msg format error)

ldoktor (approved):
lgtm, thanks

PR #150, opened 2025-05-21
Pass both ops linux/windows modules through black linting

pevogam:
This standardizes the string quote use and fixes some missed fstrings, then also formats some line lengths and other small changes.


aexpect/ops_linux.py
```diff
@@ -297,8 +301,11 @@ def copy(session, source, target, quote_path=True, flags=""):
     Calls `cp source target` through a session. See `man cp` for what source
     and target can be and what behavior to expect.
     """
-    cmd = f'cp {flags} {quote(source)} {quote(target)}' if quote_path \
-        else f'cp {flags} {source} {target}'
+    cmd = (
+        f"cp {flags} {quote(source)} {quote(target)}"
+        if quote_path
+        else f"cp {flags} {source} {target}"
+    )
```
ldoktor, line 308: Well this is a bit iffy, would you mind changing it to `if/then/else` instead?

aexpect/ops_linux.py
```diff
@@ -486,15 +493,16 @@ def hash_file(session, filename, size="", method="md5"):
...
     hash_str = output.split(maxsplit=1)[0].lower()
 
     # check that all chars are hex
-    if hash_str.strip('0123456789abcdef'):
-        raise RuntimeError('Resulting hash string has unexpected characters: '
-                           + hash_str)
+    if hash_str.strip("0123456789abcdef"):
+        raise RuntimeError(
+            f"Resulting hash string has unexpected characters: {hash_str}"
+        )
```
ldoktor, line 505: How about:

```python
raise RuntimeError("Resulting hash string has unexpected"
                   f"characters: {hash_str}")
```

ldoktor (approved):
Would you please change the `"foo" if  XX else "` to `if/then/else`? Apart from that it's fine, but perhaps we should include the black check in `make check`, what do you think?

aexpect/ops_linux.py
pevogam: I see, ok will adapt it manually then.

PR #152, opened 2025-05-26
Enable black formatter

ldoktor:
This PR is built on top of #151 and adds python black style fixes.


ldoktor:
@pevogam this should take care of black ;-)

ldoktor:
Sure, I'm not saying it's bad, it's just something I need to get used to :-)

Issue #154, opened 2025-06-06
(wiki contribute) The difference between `run_fg()` `run_bg()` `run_tail()`


AClon314:
Due to lack of development documentation, I spend some time to figure out the difference between `run_**()` in source code. I want to share my shallow point of views here, and may need re-correct. I'm honoured if these could be added in wiki page.

## run
- run_fg(): return (`int`: exit status, `str`: mixture of stdout and stderr), will auto `kill()` internally after `timeout`(default 1 seconds)
- run_bg(): interact with stdin, need **manually** `kill()` before call `get_status()`
- run_tail(): no interact, no stdin, long running program in bg, also need `kill()` before `get_status()`
- Spawn class: you have and need full manual control of the process.

## kill & close
- kill(): call this before `get_status()` to avoid endless waiting. The `get_status()` don't follow the timeout that have been previous set.
- close(): kill() and cleanup. But this would cause the later `get_status()`  always return `None`.

## pytest
```python
import pytest
import aexpect

@pytest.mark.parametrize(
    'cmds, func',
    [
        # (['echo "Hello!"'] * 2, run_fg),
        # (['mamba  env list --json', 'mamba info --json'], run_fg),
        (['aria2c --enable-rpc --rpc-listen-port=16800'] * 2, run_fg),
        # (['curl -I https://www.bing.com'] * 2, run_bg),
    ]
)
def test_aexpect(cmds: list[str], func):
    aexpects: list['aexpect.Expect' | tuple[int, str]] = [func(cmd) for cmd in cmds]
    failed = []
    while aexpects:
        p = aexpects.pop(0)
        if isinstance(p, (aexpect.Expect, aexpect.Spawn)):
            p.kill()
            failed.append(p.command) if p.get_status() != 0 else None
        elif p[0] != 0:
            failed.append(p[1])
    assert not failed, f"status != 0: {failed}"
```

ldoktor:
Hello @AClon314, thanks for your snippet. We wanted to create documentation but never allocated enough time as this library used to be part of `virttest` and `avocado-vt` projects so people simply knew how to use it. We do pay a lot of attention to docstrings, though. Perhaps we could add the `run_*` commands to [README](https://github.com/avocado-framework/aexpect/blob/main/README.rst) in a simplified way, we already have a `Simple usage` section there so we could change it to `session usage` and `run commands usage`, would that help?

[...]

PR #155, opened 2025-06-11
Docs improvements

ldoktor:
Few docs changes inspired by https://github.com/avocado-framework/aexpect/issues/154

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,6 +59,12 @@ Simple usage
     >>> session.is_responsive()
     False
 
+To get more information use the python `help()` command on various objects,
+we do keep our docstrings updated. To get details about all `aexpect.client`
+objects simply run `python -c "import aexpect; [print(f'\n{name}\n{len(name)
+* '='}', func.__doc__) for name, func in aexpect.__dict__.items() if
+hasattr(func, '__doc__')]"`
+
 Debugging
 ---------
 
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -1716,7 +1716,8 @@ def run_tail(
     encoding=None,
 ):
     """
-    Run a subprocess in the background and collect its output and exit status.
+    Run a command then await for it to finish up to timeout and return the
+    Tail object (alive or dead, allows reading output, no stdin interaction).
 
     Run command as a subprocess.  Call output_func with each line of output
     from the subprocess (prefixed by output_prefix).  Call termination_func
@@ -1769,7 +1770,8 @@ def run_bg(
     encoding=None,
 ):
     """
-    Run a subprocess in the background and collect its output and exit status.
+    Run a command then await for it to finish up to timeout and return the
+    Expect object (alive or dead, allows interaction).
 
     Run command as a subprocess.  Call output_func with each line of output
     from the subprocess (prefixed by output_prefix).  Call termination_func


ldoktor:
@AClon314 would this work for you or would you prefer another formulation?

README.rst
```diff
@@ -59,6 +59,12 @@ Simple usage
     >>> session.is_responsive()
     False
 
+To get more information use the python `help()` command on various objects,
+we do keep our docstrings updated. To get details about all `aexpect.client`
+objects simply run `python -c "import aexpect; [print(f'\n{name}\n{len(name)
+* '='}', func.__doc__) for name, func in aexpect.__dict__.items() if
+hasattr(func, '__doc__')]"`
```
pevogam, line 66: Seems a bit far fetched but good it is a one-liner.

aexpect/client.py
```diff
@@ -1716,7 +1716,8 @@ def run_tail(
     encoding=None,
 ):
     """
-    Run a subprocess in the background and collect its output and exit status.
+    Runs the command, awaits for it to finish up to timeout and returns the
```
pevogam, line 1719: The default for docstrings is to use imperative form just like commit message. So I think `Run` is still better than `Runs`.

pevogam: In fact it would even be better as "Run a command" instead of "Run the command" since the argument is a detail of the function and not the other way around.

pevogam: Run a command then wait for it to finish up to a given timeout...

aexpect/client.py
```diff
@@ -1769,7 +1770,8 @@ def run_bg(
...
+    Expect object (alive or dead, allows interaction).
```
pevogam, line 1774: Same as above - of course nice clarification too, perhaps we should take it out of the brackets as it is not secondary clarification but a major piece of information to separate the two functions.

pevogam (commented):
I left a quick review about this

aexpect/client.py
ldoktor: ok

aexpect/client.py
ldoktor: ok, not sure how to formulate that better without adding long sentences, so let's just go with the brackets form for now...

README.rst
```diff
@@ -59,6 +59,12 @@ Simple usage
     >>> session.is_responsive()
     False
 
+To get more information use the python `help()` command on various objects,
+we do keep our docstrings updated. To get details about all `aexpect.client`
+objects simply run `python -c "import aexpect; [print(f'\n{name}\n{len(name)
```
wainersm, line 64: Hi @ldoktor ! do you expect it to describe internal objects such as `__name__`? Maybe they aren't so relevant and can be filtered out? If too much a work then ignore that comment.

Ran locally just fine:

```shell
$ python -c "import aexpect; [print(f'\n{name}\n{len(name)
* '='}', func.__doc__) for name, func in aexpect.__dict__.items() if
hasattr(func, '__doc__')]"

__name__
======== str(object='') -> str
str(bytes_or_buffer[, encoding[, errors]]) -> str

Create a new string object from the given object. If encoding or
errors is specified, then the object must expose a data buffer
that will be decoded using the given encoding and error handler.
Otherwise, returns the result of object.__str__() (if defined)
or repr(object).
encoding defaults to 'utf-8'.
errors defaults to 'strict'.

<snip>

remote
====== 
Functions and classes used for logging into guests and transferring files.

Example usage:

::

    import aexpect
    import aexpect.remote

    with aexpect.ShellSession("sh") as b:
        b.cmd("echo asdf > /tmp/bbb")

    with aexpect.remote.remote_login("ssh", "127.0.0.1", 5000, "root", "123456",
                                    aexpect.remote.PROMPT_LINUX) as a:
        a.cmd("rm /tmp/aaa -Rf")
        aexpect.remote.scp_to_session(a, "/tmp/bbb", "/tmp/aaa")
        aexpect.remote.scp_from_session(a, "/tmp/aaa", "/tmp/ccc")

.. warning:: Some of this API is freshly migrated and might still be experimental.
```
```

ldoktor: TBH I don't think it'd be worth complicating the command (as it's suppose to be executed by users so they should understand that before running it)

wainersm: Fair enough!

wainersm (approved):

PR #147, opened 2025-04-30
Failed kill

ldoktor:
@smitterl identified a bug in aexpect, when unkillable `aexpect_worker` results in indefinite hang of `Spawn.close()`. To mitigate that I added a timeout to our `wait_for_lock` and used it in `close()` (this fix can be tested by commenting-out the `kill`). Then I noticed the `close()` can actually be called multiple times when the user uses multiple processes/threads. To address that I'd suggest moving to BSD flock rather than posix lockf as that one protects individual threads and adding another `close` lock file to protect this critical path. This can be tested by:

```
import aexpect
import threading
#import pydevd

def close_session(session):
    """Closes the given session."""
    #pydevd.settrace("127.0.0.1")
    session.close()

# Assuming you have a list of sessions to be closed
session = aexpect.ShellSession("bash")

threads = []
for _ in range(20000):
    thread = threading.Thread(target=close_session, args=(session,))
    threads.append(thread)

for thread in threads:
    thread.start()

for t in threads:
    t.join()

del(session)

print("DONE")
```

One thing I'm not completely sure about is whether we don't rely on the non-thread-safe lockf in aexpect so please test this thoroughly.

PR #156, opened 2025-06-18
client: Avoid double-cleanup exception

ldoktor:
in python <3.13 it raises FileNotFoundError, but python 3.13 seems to raise OSError exception instead when trying to unlock non-existing file.

Fixes: https://github.com/avocado-framework/aexpect/pull/147#issuecomment-2970833737

### DIFF
--- a/aexpect/client.py
+++ b/aexpect/client.py
@@ -473,7 +473,7 @@ class Spawn:
                 try:
                     unlock_fd(lock)
                     os.unlink(self._close_lockfile)
-                except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


ldoktor:
@pevogam if I read the traceback correctly it seems python3.13 raises a different exception when the file was already deleted (forceful cleanup or multiple cleanup). Please let me know if there are any other issues as once this is resolved I'd like make a release as there were quite a few important changes already.

pevogam (approved):
Indeed, affected tests pass now and this change is easy enough to review and say yes to.

PR #157, opened 2025-07-29
aexpect: Version 1.8.0

ldoktor:
Fixes: #127


pevogam:
None at all :tada:

pevogam:
Ah false alarm, this somehow worked before but now pulled the wrong aexpect 1.7.0 which is incompatible with Python 3.13. Sorry for the noise!

ldoktor:
Uff, thanks for the confirmation :-)

PR #163, opened 2026-01-02
Emit a helpful logging message on reused object server

pevogam:
## Summary by CodeRabbit

* **Chores**
  * Enhanced logging to explicitly note when an existing remote object server connection is reused after binding. Improves diagnostics and operational visibility for troubleshooting and monitoring. This change is purely informational and does not alter behavior or control flow.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/aexpect/remote_door.py
+++ b/aexpect/remote_door.py
@@ -626,6 +626,7 @@ def get_remote_object(
         remote_object = Pyro4.Proxy(f"PYRONAME:{object_name}@{host}:{port}")
         # noinspection PyProtectedMember
         remote_object._pyroBind()
+        LOG.warning("Previous remote object server running, reusing it as is")
     except Pyro4.errors.PyroError as error:
         if not session:
             raise


aexpect/remote_door.py
```diff
@@ -626,6 +626,7 @@ def get_remote_object(
         remote_object = Pyro4.Proxy(f"PYRONAME:{object_name}@{host}:{port}")
         # noinspection PyProtectedMember
         remote_object._pyroBind()
+        LOG.info("Previous remote object server running, reusing it as is")
```
ldoktor: Wouldn't `debug` loglevel suffice here?

pevogam: Hmmm, I was even considering warning since this tripped me quite a few times in the past. The log message points out a caveat so higher visibility is more advisable here.

ldoktor: If you consider this very important then warning would suit the best.

pevogam: Alright, let's do this then, especially since reusing a previous running remote object server is a less standard use case.

ldoktor (approved):
Thanks

PR #164, opened 2026-01-02
Disable pseudo-terminal allocation for the rsync copy backend

pevogam:
In cases of hanging aexpect processes it is possible that we exhaust the pseudo-terminals on the system. Disabling pseudo-terminal allocation for the rsync backend should help mitigate this issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated SSH configuration in remote file synchronization operations.

* **Documentation**
  * Revised remote file transfer examples to reflect current practices.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -667,7 +667,7 @@ def remote_copy(
     Transfer files using rsync or SCP, given a command line.
 
     :param command: The command to execute
-        (e.g. "rsync -avz -e 'ssh -p 22' /local/path user@host:/remote/path"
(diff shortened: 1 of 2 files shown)


ldoktor (approved):
Thanks, should be better to use -T for rsync anyway. I have tested it with interactive password and it seemed to work well as well (I was afraid it might not work).

pevogam:
Integration tested on my side as well in our avocado test suite so verified to work before I pushed. Thanks for the quick reaction!

PR #167, opened 2026-05-29
Allow retries for all major remote copy methods

### LINKED ISSUE #166: Retriable copies for the remote module
In some cases of network instability (e.g. DNS server reset or similar) if aexpect is in use this will result in possibly job crashing errors like
```
  File "/usr/lib/python3.13/site-packages/aexpect/remote_door.py", line 308, in run_subcontrol
    remote_control_path = _copy_control(session, control_path)

  File "/usr/lib/python3.13/site-packages/aexpect/remote_door.py", line 283, in _copy_control
    remote.copy_files_to(
    ~~~~~~~~~~~~~~~~~~~~^
        session.host,
        ^^^^^^^^^^^^^
    ...<5 lines>...
        remote_control_path,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^

  File "/usr/lib/python3.13/site-packages/aexpect/remote.py", line 1378, in transfer
    ret = func(*args, **kwargs)

  File "/usr/lib/python3.13/site-packages/aexpect/remote.py", line 1432, in copy_files_to
    scp_to_remote(
    ~~~~~~~~~~~~~^
        address,
        ^^^^^^^^
    ...<10 lines>...
        interface=interface,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^

  File "/usr/lib/python3.13/site-packages/aexpect/remote.py", line 755, in scp_to_remote
    return remote_copy(
        command, password_list, log_filename, log_function, timeout
    )

  File "/usr/lib/python3.13/site-packages/aexpect/remote.py", line 696, in remote_copy
    _remote_copy(
    ~~~~~~~~~~~~^
        session, password_list, transfer_timeout, login_timeout, method
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^

  File "/usr/lib/python3.13/site-packages/aexpect/remote.py", line 646, in _remote_copy
    raise SCPError("SCP client said 'lost connection'", text)
```
I think it might be a good idea if the remote copy operations support a number of (re)tries (with default `tries=1`). This will make using aexpect's remote layer in automation more resilient.

pevogam:
Resolves https://github.com/avocado-framework/aexpect/issues/166.

### DIFF
--- a/aexpect/remote.py
+++ b/aexpect/remote.py
@@ -1107,6 +1107,7 @@ def nc_copy_between_remotes(
     s_session=None,
     d_session=None,
     file_transfer_timeout=600,
+    tries=1,
 ):
     """
     Copy files from guest to guest using netcat.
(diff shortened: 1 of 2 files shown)


pevogam:
Hi @ldoktor, if you are overloaded I can always join in here and help out with the project. Just let me know.

aexpect/remote.py
```diff
@@ -677,25 +678,67 @@ def remote_copy(
...
+                    attempt + 1,
+                    tries,
+                    error,
+                )
+                time.sleep(1)  # small delay before retry
+            else:
+                raise
+        except (TransferFailedError, SCPError, RsyncError) as error:
+            # For transfer failures, only retry on specific conditions
+            if "Connection" in str(error) or "timeout" in str(error).lower():
```
ldoktor: Parsing error string can be fragile, I'd suggest using `lower` for both (ideally convert it at the beginning to avoid double conversion)

aexpect/remote.py
```diff
@@ -662,6 +662,7 @@ def remote_copy(
     log_function=None,
     transfer_timeout=600,
     login_timeout=300,
+    tries=1,
```
ldoktor: I'd suggest using `attempts` to be consistent with log messages as well as other parts of avocado

aexpect/remote.py
```diff
@@ -677,25 +678,67 @@ def remote_copy(
...
-        output_params = ()
     method = "rsync" if "rsync" in command else "scp"
-    with Expect(
-        command, output_func=output_func, output_params=output_params
-    ) as session:
-        _remote_copy(
-            session, password_list, transfer_timeout, login_timeout, method
-        )
+
+    for attempt in range(tries):
```
ldoktor: How about `range(1, tries + 1)` to avoid all the `+1` handling?

aexpect/remote.py
```diff
@@ -1132,59 +1133,91 @@ def nc_copy_between_remotes(
...
+                LOG.debug(
+                    "Transfer failed on attempt %d/%d, retrying: %s",
+                    attempt + 1,
+                    tries,
+                    error,
+                )
+                time.sleep(1)  # small delay before retry
+                # reset sessions for retry
+                if s_session:
+                    s_session.close()
```
ldoktor: This closes even connections that might have been passed by the user as an argument, which might not be expected and can break existing code which might expect the session intact.

aexpect/remote.py
```diff
@@ -1302,23 +1336,43 @@ def stop_server(session):
...
+                        f"file {s_path} md5sum is '{src_md5}', "
+                        f"but the file {d_path} md5sum is {dst_md5}"
+                    )
+                    raise UDPError(err_msg)
+            finally:
+                stop_server(s_session)
+                s_session.close()
+                d_session.close()
+            return  # transfer is successful
+        except UDPError as error:
```
ldoktor, line 1384: Shouldn't this also catch the possible `LoginTimeoutError` from the `remote_login` (lines 1341 and 1344)?

aexpect/remote.py
```diff
@@ -1302,23 +1336,43 @@ def stop_server(session):
...
-        start_client(d_session)
-        dst_md5 = get_file_md5(d_session, d_path)
-        if src_md5 != dst_md5:
-            err_msg = (
-                "Files md5sum mismatch, "
-                f"file {s_path} md5sum is '{src_md5}', "
-                f"but the file {d_path} md5sum is {dst_md5}"
+    for attempt in range(tries):
+        try:
+            s_session = remote_login(
```
ldoktor: In case the `d_session`'s `remote_login` fails the `s_session` is not cleaned. How about setting it to `None` and moving the actual logins to the `try/finally` block?

ldoktor (commented):
Thanks @pevogam, the re-try mechanism should help a lot. There are few concerns but otherwise it should do.

And I'm sorry about the delays, I took a series of PTOs recently, hopefully it'll be better now.

aexpect/remote.py
pevogam: The sessions are entirely recreated on each try yes, especially because stale sessions might become unrecoverable and no retry would be able to recover the copy process. Perhaps if we detect the arguments were passed we should simply disable the retry behavior, e.g. set the `attempts=1`?

aexpect/remote.py
pevogam: I don't consider the login error as one we should catch here, it is after all meant to only retry on specific failures of the transfer process. If we fail to obtain a new session (login) then this should also be fatal towards the retries as it indicates a deeper problem. Do you agree?

pevogam: I guess we could just optionally close if the args are not None, let's see

aexpect/remote.py
pevogam: Yes, make sense - added multiple try/finally blocks though to make the sequence guaranteed.

aexpect/remote.py
pevogam: I guess we could just optionally close if the args are not None, let's see

pevogam:
I pushed a second commit that now actually improves previous RSS client code in terms of narrowed exception handling to address the latest autoreview. Other than that I think everything else was address and we would definitely benefit from all additional test coverage.

ldoktor (approved):
Thank you, lgtm

PR #165, opened 2026-01-02
Restore RPM builds from pyproject configuration

pevogam:
## Summary by CodeRabbit

* **Chores**
  * Updated project version to 1.8.0
  * Refined project configuration and build system setup

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #172, opened 2026-07-21
Fix "make rpm"

ldoktor:
this is a follow-up after https://github.com/avocado-framework/aexpect/pull/165 which fixes the Makefile so the "make rpm" works and then contains extra commit to attempt to restore dynamic versioning for pip versions while keeping hardcoded one for RPMs.

@pevogam please try this out I hope it'll work for you as well as for me. My main problem with the current version is that any issue reported will only contain the hardcoded version but we will not know whether it was released, development or even custom dirty version. I'd really like to have some dynamic versioning in-place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PYTHON=$(shell which python 2>/dev/null || which python3 2>/dev/null)
 PROJECT=aexpect
-VERSION=$(shell $(PYTHON) -m setuptools_scm)
+VERSION=$(shell $(PYTHON) -m setuptools_scm 2>/dev/null || grep '^Version:' python-$(PROJECT).spec | awk '{print $$2}')
+SPEC_VERSION=$(shell grep '^Version:' python-$(PROJECT).spec | awk '{print $$2}')
 COMMIT=$(shell git log --pretty=format:'%H' -n 1)
 SHORT_COMMIT=$(shell git log --pretty=format:'%h' -n 1)
 COMMIT_DATE=$(shell git log --pretty='format:%cd' --date='format:%Y%m%d' -n 1)
@@ -25,7 +26,7 @@ source: clean
 
 source-release: clean
 	mkdir -p SOURCES
-	git archive --prefix="$(PROJECT)-$(VERSION)/" -o "SOURCES/$(PROJECT)-$(VERSION).tar.gz" $(VERSION)
+	git archive --prefix="$(PROJECT)-$(SPEC_VERSION)/" -o "SOURCES/$(PROJECT)-$(SPEC_VERSION).tar.gz" $(SPEC_VERSION)
 
 install:
 	rm -r dist 2>/dev/null || true
@@ -74,7 +75,7 @@ srpm: source
 
 rpm: srpm
 	mkdir -p BUILD/RPM
-	mock -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "rel_build 0" -D "commit $(COMMIT)" -D "commit_date $(COMMIT_DATE)" --rebuild BUILD/SRPM/python-$(PROJECT)-$(VERSION)-*.src.rpm
+	mock -r $(MOCK_CONFIG) --resultdir BUILD/RPM -D "rel_build 0" -D "commit $(COMMIT)" -D "commit_date $(COMMIT_DATE)" --rebuild BUILD/SRPM/python-$(PROJECT)-$(SPEC_VERSION)-*.src.rpm
 
 srpm-release: source-release
 	mkdir -p BUILD/SRPM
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 PYTHON=$(shell which python 2>/dev/null || which python3 2>/dev/null)
 PROJECT=aexpect
-VERSION=$(shell $(PYTHON) -m setuptools_scm)
+VERSION=$(shell $(PYTHON) -m setuptools_scm 2>/dev/null || grep '^Version:' python-$(PROJECT).spec | awk '{print $$2}')
+SPEC_VERSION=$(shell grep '^Version:' python-$(PROJECT).spec | awk '{print $$2}')
```
pevogam, line 4: This line seems repetitive of the one above, how about simply getting SPEC_VERSION first and then retrieving the backup version directly from this variable being already set?

Makefile
```diff
@@ -25,7 +26,7 @@ source: clean
 
 source-release: clean
 	mkdir -p SOURCES
-	git archive --prefix="$(PROJECT)-$(VERSION)/" -o "SOURCES/$(PROJECT)-$(VERSION).tar.gz" $(VERSION)
+	git archive --prefix="$(PROJECT)-$(SPEC_VERSION)/" -o "SOURCES/$(PROJECT)-$(SPEC_VERSION).tar.gz" $(SPEC_VERSION)
```
pevogam, line 29: So I assume this and the below diff blocks need SPEC_VERSION explicitly? What would happen if the SPEC_VERSION differs from the VERSION here? Would this still be reasonable?

pyproject.toml
```diff
@@ -44,6 +44,7 @@ include = ["aexpect*"]
 [tool.setuptools_scm]
 version_scheme = "post-release"
 local_scheme = "node-and-date"
+fallback_version = "1.8.0"
```
pevogam, line 47: So in some sense we still end up with non-dynamic version, no? Because either we have to update this each time as we would have updated `version` or we don't update it but get strange cases where higher version might display as lower version which could cause some havoc in deployment pipelines.

pevogam (commented):
Hi @ldoktor, the build indeed passes just fine. However I have one major question in the comments below in addition to a few minor change suggestions.

Makefile
ldoktor: sure

Makefile
ldoktor: The previous (hardcoded) version used a fixed version to avoid problems passing it between mock and the repo. To keep this behaviour I added the SPEC_VERSION which is used for `rpm` targets while keeping it dynamic for the other targets. So different version should result in the RPM force-setting to the SPEC one.

Ideally I'd like to remove the `SPEC_VERSION` completely, but you claimed it didn't work on your build system.

pevogam: > The previous (hardcoded) version used a fixed version to avoid problems passing it between mock and the repo. To keep this behaviour I added the SPEC_VERSION which is used for rpm targets while keeping it dynamic for the other targets. So different version should result in the RPM force-setting to the SPEC one.

Meaning that we still have to maintain at least one hard-coded version or somewhat worse yet we may want to ignore it and only rely on the dynamical version in all other cases, ending up with diverging versions. Of course one may argue that manual version maintenance is now restricted to just people caring about the RPM deployment but at the same time it does sound like it will be much more tempting for everyone else to just ignore updating it and for the project to end up with different python versions.

> Ideally I'd like to remove the SPEC_VERSION completely, but you claimed it didn't work on your build system.

If you want you could suggest a more daring dynamic version and I could still try it. I am performing my integration testing in a fairly isolated Fedora 42+ LXC container without adding much complexity to such testing environment beyond the goal of running avocado plugins.

python-aexpect.spec
```diff
@@ -60,6 +62,7 @@ sftp, telnet, among others.
 %endif
 
 %build
+export SETUPTOOLS_SCM_PRETEND_VERSION=%{version}
```
pevogam, line 65: For some comparison, here is the diff with my draft that failed to achieve a build so far:
```diff
---
 aexpect/__init__.py | 2 ++
 pyproject.toml      | 1 +
 python-aexpect.spec | 7 +++++++
 3 files changed, 10 insertions(+)

diff --git a/aexpect/__init__.py b/aexpect/__init__.py
index 3e32262..6d714d1 100644
--- a/aexpect/__init__.py
+++ b/aexpect/__init__.py
@@ -14,6 +14,8 @@ Aexpect module, see help('aexpect.client') to get info about the main
 entry-points.
 """

+# _version is generated by tool.setuptools_scm
+from ._version import version as __version__
 from . import remote, rss_client
 from .client import (
     Expect,
diff --git a/pyproject.toml b/pyproject.toml
index 3975c3d..dfb3782 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ include = ["aexpect*"]
 [tool.setuptools_scm]
 version_scheme = "post-release"
 local_scheme = "node-and-date"
+write_to = "aexpect/_version.py"

 [tool.black]
 line-length = 79
diff --git a/python-aexpect.spec b/python-aexpect.spec
index 955ef09..0066d36 100644
--- a/python-aexpect.spec
+++ b/python-aexpect.spec
@@ -36,8 +36,10 @@ Source0: %{url}/archive/%{commit}/%{gittar}
 %endif

 BuildArch: noarch
+BuildRequires: pyproject-rpm-macros
 BuildRequires: python3-devel
 BuildRequires: python3-setuptools
+BuildRequires: python3-setuptools-scm

 %description
 Aexpect is a python library used to control interactive applications, very
@@ -59,7 +61,12 @@ sftp, telnet, among others.
 %autosetup -n aexpect-%{commit} -p 1
 %endif

+%generate_buildrequires
+export SETUPTOOLS_SCM_PRETEND_VERSION=%{version}
+%pyproject_buildrequires
+
 %build
+export SETUPTOOLS_SCM_PRETEND_VERSION=%{version}
 %pyproject_wheel

 %install
```

